PR #5, opened 2021-04-21
feat: fiddle remark plugin

erickzhao:
This PR adds a remark plugin that transforms fenced code blocks into a tabbed editor view, with a link to open the example in Electron Fiddle.

Syntax:
````md
 ```fiddle docs/fiddles/quickstart
 ```
````

Demo:

https://user-images.githubusercontent.com/16010076/115620084-951fe400-a2a9-11eb-9e58-898d84b36a28.mp4

### CODE REVIEW

File: src/transformers/fiddle-embedder.js [hunk 1]
```diff
@@ -0,0 +1,147 @@
...
+      });
+    } else {
+      children.push(
+        {
+          type: 'jsx',
+          value: `</TabItem>\n</Tabs>`,
+        },
+        {
+          type: 'jsx',
+          value: `<LaunchButton url="https://fiddle.electronjs.org/launch?target=electron/v12.0.4/${dir}"/>`
```
Ethan-Arrowood, line 139: Is the version part of this string configurable somehow? Seems like this would miss updates whenever new versions are released. Maybe it could reference the package json or something?

Ethan-Arrowood (approved):
LGTM. Comment is non-blocking, just a consideration

src/transformers/fiddle-embedder.js
erickzhao: Yep, this is just part of the prototype. I'll move the version out to a variable and add a TODO so that it's less hard-coded!

molant (approved):

PR #20, opened 2021-04-29
chore: automatically use the latest stable release

molant:


src/transformers/fiddle-embedder.js
```diff
@@ -2,9 +2,16 @@
 const visitParents = require('unist-util-visit-parents');
 const path = require('path');
 const fs = require('fs-extra');
+const latestVersion = require('latest-version');
 
-//TODO: Make this the latest stable version instead of hardcoding
-const VERSION = 'v12.0.4';
+let _version = '';
```
erickzhao, line 7: Is `_version` to cache the version tag? Why is this needed between runs?

molant: I'm assuming that transformer is called multiple times in the build process (haven't dig the docusaurus doc to verify this though).
This way we do only one network request per `require` instead of one per call.

erickzhao: I tried playing around with log statements and that indeed seems to be the case. Makes sense!

erickzhao (approved):

Issue #2, opened 2021-04-16
[☂] v1 tasks


molant:
Some tasks from the top of my head:

Pending:

- [ ] Figure out i18n integration (#64)
- [ ] Figure out docs versioning
- [ ] Review and update documentation style guide to match new website.
- [ ] Create new docs landing page (@molant, @erickzhao)
- [ ] Upstream everything we can to `electron/electron`
- [x] Update styles to match a bit more the current live site (@erickzhao)
- [ ] Need to find a way to allow easy merges on `electron/electron` for the `/docs`  folder ➡ Maybe use the CODEOWNERS plugin or similar?

Done:

- [x] Add frontmatter to documentation, should be in electron/electron but maybe we can start doing it here (see #https://github.com/electron/electron/pull/28303)? (@molant) (#3)
- [x] Pull documentation from the right branch as pre-build step (@molant) (#3)
- [x] Automatically generate the sidebar ➡ ~~Haven't found any documentation about it or if it is enough with the right folder structure~~ Looks like we need to generate it somehow or else docusaurus complains with:  (@molant) (#3)
  > Error: Bad sidebars file.
  > These sidebar document ids do not exist:
- [x] Add Fiddle integration in the docs (@erickzhao)
- [x] Remove Blog link for now (#10)
- [x] Deploy the website somewhere (@molant) ➡️  https://beta.electronjs.org/
- [x] Reach out to docsearch see if we can have 2 indexes for the website (@molant) (#22) ➡ **Waiting for next crawl before merging*
- [x] Figure out how to enable fast docs development locally (@molant) (#4)
- [x] List of examples sooner rather than later
- [x] Automatically update/deploy periodically or when there's new content in `electron/electron` (#19 - @molant)
- [x] Add an "Examples"  link to the top nav ➡ Talk about fiddle in the landing page (@molant)
- [x] search feature
- [x] Update electronjs.org-new README
- [x] Document and be clear about acceptance criteria for submissions.
- [x] START HERE URL
- [x] Discord Channel `#TBD`

molant:
Looks like we need to generate it somehow or else docusaurus complains with: 
  > Error: Bad sidebars file.
  > These sidebar document ids do not exist:

molant:
- [ ] Remove Blog link for now
- [ ] Add an "Examples"  link to the top nav

molant:
@erickzhao I've reorganized a bit the top comment and added some new tasks.
[...]

PR #22, opened 2021-05-03
feat: enable docsearch

molant:
Enable docsearch module with provided configuration by algolia.

Currently the search does not seem to work when `contextSearch` is
enabled. This seems to be a temporary issue per Clement's mail:

> As the DocSearch plugin is not active yet, you won't be able to use
the Docusaurus `contextualSearch` option until next crawl.

While this feature is not necessary (there are no versioned docs),
it seems like the search results page has it enabled by default
always ([1]) so might as well have it enabled everywhere to make sure it
works.

[1]: https://github.com/temporalio/documentation/pull/357/files#r616656533
Ref: https://github.com/facebook/docusaurus/issues/3805
Part of #2

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -79,6 +79,11 @@ module.exports = {
       ],
       copyright: `Copyright © ${new Date().getFullYear()} My Project, Inc. Built with Docusaurus.`,
     },
+    algolia: {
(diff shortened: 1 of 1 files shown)


molant:
Before merging this we need to make sure that `contextualSearch` works so we will have to wait ~24h and deploy the site somewhere. @erickzhao is there a way we can enable "quick test deploys" from GitHub's UI like in the main site repo?

PR #36, opened 2021-05-20
feat: allow different files to be focused in an embedded fiddle

tonyferrell:
Add support for custom parameterization in the fiddle's code block header.

Example:
````
```javascript fiddle='docs/fiddles/features/keyboard-shortcuts/web-apis|focus=renderer.js'
```
````

This will default the focused tab to `renderer.js`.

image

To minimize focus issues as fiddles are changed, if the provided file is not found a compile-time error will occur.

image

Fixes: #34
cc: @erickzhao

### DIFF
--- a/src/transformers/fiddle-embedder.js
+++ b/src/transformers/fiddle-embedder.js
@@ -59,7 +59,7 @@ async function transformer(tree) {
     const options = {};
 
     // If there are optional parameters, parse them out to pass to the getFiddleAST method.
-    if (others.length) {
+    if (others.length > 0) {
       for (const option of others) {
         // Use indexOf to support bizzare combinations like `|key=Myvalue=2` (which will properly 
         // parse to {'key': 'Myvalue=2'})
--- a/src/transformers/fiddle-embedder.js
+++ b/src/transformers/fiddle-embedder.js
@@ -53,12 +53,27 @@ async function transformer(tree) {
    */
   function visitor(node, ancestors) {
     const parent = ancestors[0];
-    const folder = node.meta;
+    // Supported formats are fiddle='<folder>|<option>|option'
+    // Options must be of the format key=value with no additional quotes.
+    const [folder, ...others] = node.meta.split("|");
+    const options = {};
+
+    // If there are optional parameters, parse them out to pass to the getFiddleAST method.
+    if (others.length) {
+      for (const option of others) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/transformers/fiddle-embedder.js [hunk 1]
```diff
@@ -53,12 +53,27 @@ async function transformer(tree) {
...
   function visitor(node, ancestors) {
     const parent = ancestors[0];
-    const folder = node.meta;
+    // Supported formats are fiddle='<folder>|<option>|option'
+    // Options must be of the format key=value with no additional quotes.
+    const [folder, ...others] = node.meta.split("|");
+    const options = {};
+
+    // If there are optional parameters, parse them out to pass to the getFiddleAST method.
+    if (others.length) {
```
Ethan-Arrowood, line 62:
```suggestion
    if (others.length > 0) {
```
I prefer a more deliberate check here

src/transformers/fiddle-embedder.js
```diff
@@ -84,6 +99,10 @@ function getFiddleAST(dir, version) {
     return children;
   }
 
+  if (!fileNames.includes(focus)) {
+    throw new Error(`Provided focus (${focus}) is not an available file in this fiddle (${dir}). Available files are [${fileNames.join(",")}]`);
```
erickzhao, line 103: nit: slightly more human-readable

```suggestion
    throw new Error(`Provided focus (${focus}) is not an available file in this fiddle (${dir}). Available files are [${fileNames.join(", ")}]`);
```

erickzhao (approved):

src/transformers/fiddle-embedder.js
erickzhao: I think Ethan's suggestion makes it more explicit, which is good.

erickzhao (changes requested):
One comment, actually: this will break backward compatibility with `electronjs.org`. 

Can you filter these parameters out in a separate PR in https://github.com/electron/i18n/blob/master/lib/transfomers/remark-fiddle-urls.ts ?

tonyferrell:
> One comment, actually: this will break backward compatibility with `electronjs.org`.
> 
> Can you filter these parameters out in a separate PR in https://github.com/electron/i18n/blob/master/lib/transfomers/remark-fiddle-urls.ts ?

See electron/i18n/pull/1882

erickzhao (approved):

PR #41, opened 2021-06-03
chore: update default path to /docs/latest/

molant:
Adjust scripts to take into account the new path for the documentation
so we can be future proof if we ever decide to have versioning.


molant:
I'm sure there's a prettier way to do this but we can probably postpone this for when we do versioning.

erickzhao (commented):
Seems like there's a deploy failure

molant:
@erickzhao deployment is finally green!

molant:
This is obsolete after #42 got merged 🎉

PR #59, opened 2021-06-22
feat: improve footer content

### LINKED ISSUE #51: Organize footer content

erickzhao:
Closes #51 

Adds the @openjs-foundation logo to the new docs site, updates the copyright info, and reorganizes the footer links


docusaurus.config.js
```diff
@@ -63,22 +68,14 @@ module.exports = {
               to: '/',
             },
             {
-              label: 'Performance',
-              to: '/performance/',
-            },
-            {
-              label: 'Security',
-              to: '/security/',
+              label: 'API Reference',
```
molant, line 71: What do you think of bubbling up the perf and security links somehow?
From the latest survey they were a lot of devs that weren't aware of these lists and we want to make sure they have them in mind.

erickzhao: @molant I feel like bubbling it up somehow is a good idea, but the current footer content felt a bit jarring since these made up 2/3 items in the Docs section 🤔

molant: We can always add more links if we think those are relevant.
My main concern is that we are duplicating the links that are already in the header that are already visible and from what I'm seeing in the stats the perf one is driving some traffic.



Ideally we should add some analytics to see how many people click

erickzhao: addressed in e6ed369

i18n/en-US/code.json
```diff
@@ -39,6 +39,10 @@
     "message": "Read More",
     "description": "The label used in blog post item excerpts to link to full blog posts"
   },
+  "theme.blog.sidebar.navAriaLabel": {
```
erickzhao, line 42: note: this was generated by the pre-commit hook script. I'm not sure why we moved the `"theme.blog.sidebar.navAriaLabel"` key up a few spaces. Might be a bug?

molant: The generation of those files is unfortunately not deterministic. I had to use `--no-verify` a few times the last few days 😕

@erickzhao maybe something we could contribute upstream?

PR #66, opened 2021-07-09
feat: initial work on i18n

molant:


docusaurus.config.js
```diff
@@ -12,6 +12,10 @@ module.exports = {
   favicon: 'img/favicon.ico',
   organizationName: 'electron',
   projectName: 'electron',
+  i18n: {
+    defaultLocale: 'en',
+    locales: ['en', 'es-ES']
```
molant, line 17: Testing the Spanish one for now although all the translations are pulled and organized.
We could keep this in sync automatically with Crowdin somehow (during the prebuild step probably?)

docusaurus.config.js
```diff
@@ -99,7 +107,7 @@ module.exports = {
           ],
         },
       ],
-      copyright: `Copyright © ${new Date().getFullYear()} My Project, Inc. Built with Docusaurus.`,
+      copyright: `Copyright © OpenJS Foundation and Electron contributors.`,
```
molant, line 110: I removed the year because the generated JSON had it hardcoded somehow instead of doing interpolation and thus it will be kept out of date.

scripts/tasks/md-fixers.js
```diff
@@ -4,38 +4,6 @@ const fs = require('fs').promises;
 const path = require('path');
 const globby = require('globby');
 
-/** The keywords that need to be escaped so MDX does not complain */
-const keywords = new Set([
```
molant, line 8: This wasn't being used at the end

scripts/tasks/reorg-docs.js
```diff
@@ -0,0 +1,136 @@
+//@ts-check
```
molant, line 1: This should be mostly the previous contents of `download-docs.js`.

scripts/tasks/reorg-docs.json
```diff
@@ -0,0 +1,82 @@
+{
```
molant, line 1: This was `docs-reorg.json` that I renamed to match `reorg-docs.js`.

scripts/tasks/download-translations.js
```diff
@@ -0,0 +1,115 @@
...
+  let build;
+
+  for (let i = 0; i < counter; i++) {
+    build = await getBuild(buildId);
+    console.log(`Crowdin status: Project ${buildId} - ${build.status}`);
+
+    if (build.status === 'finished') {
+      break;
+    } else {
+      console.log(`Crowdin status: Waiting ${interval} seconds (retry ${i}/${counter})`)
```
molant, line 78: In my tests I've got builds done at around 3.5 minutes (interval 7). We could increate the counter a bit more if we want to be extra catious.

.env.example
```diff
@@ -1 +1,3 @@
-GITHUB_TOKEN=
\ No newline at end of file
+GITHUB_TOKEN=
+CROWDIN_TOKEN=
```
molant, line 2: We need to add this tokens to Heroku

molant:
This should be ready for an initial round of feedback.
The biggest blocker at the moment I think it's that the routes are `/docs/latest/es-ES`. It will probably be better for them to be `/es-ES/docs/latest/` and all be lower case.

This will touch docusaurus and CDN configurations. And we might want to consider the scenario for versioned docs as well.

Issue #64, opened 2021-07-08
Internazionalization


molant:
This is an outline of how we could handle the internationalization if the website based on our current experience in the main site and [Docusaurus documentation](https://docusaurus.io/docs/2.0.0-beta.0/i18n/introduction).

**Some considerations**

* Each language is in its own folder under `i18n/`. In our case it will be 7 languages (Chinese Simplified, French, German, Japanese, Portuguese, Russian, and Spanish)
* We will have to add a `code.json` and `navbar.json` to each [locale for translation](https://docusaurus.io/docs/2.0.0-beta.0/i18n/tutorial#translate-json-files) so the navigation and other strings are translated.
* We might need to use [explicit ids in headings](https://docusaurus.io/docs/2.0.0-beta.0/markdown-features/headings#explicit-ids) to make sure relative links still work (not sure what the current status is)
* There is no automatic locale detection (its an static site). We could use some JavaScript or show the English site first and let the users pick their preferred one via the [`localeDropdown`](https://docusaurus.io/docs/2.0.0-beta.0/i18n/tutorial#theme-configuration)

**How to?**

I think the less disruptive way would be to:

* Pull the `i18n` package periodically and move the files to the right place (at least until the changes happen upstream)
* Make `i18n` pull the required files for translations (e.g.: `navbar.json`)
* The English language keeps being updated as it is right now

molant:
I've been looking into fixing the build crashes in #78 and haven't been super successful 😞

I found [this issue](https://github.com/facebook/docusaurus/issues/4765) where using esbuild speeds the build process. I'm trying it locally and seems to speed it up, but I still can't build with all the languages in this PR and it takes forever (15 minutes until I get a JavaScript heap out of memory with 5/8 languages built).

My current plan to get this through the finish line is:
- [ ] Do a separate PR with the esbuild patch: This should improve build times for everything
- [ ] Commit docs and i18n markdown documents on this repo
- [ ] Use this repo as the source of truth for Crowdin and change how the integration is done in the PR. This will remove any dependency on i18n
- [ ] Pull Crowdin changes periodically on a GitHub action
- [ ] Build each locale independently and publish it to a different dino in Heroku, and configure Fastly to serve them as a path and not a subdomain. This is the [Hybrid approach] in Docusaurus configuration. I'm not sure if we will need to do any path transformation in Fastly or maybe specify a different path on build time.

PR #67, opened 2021-07-13
feat: add pre-commit hook to keep l10n in sync

molant:
The locale `i18n/en-US` is used by `electron/i18n` as the source to
translate content into other locales and avoid having to manually create
the different JSON files or keep them in sync.

Once those files are translated, they will be downloaded and put in
the right place for each locale. This will happen in an upcoming PR.

Rel: #64


scripts/update-l10n-sources.js
```diff
@@ -0,0 +1,33 @@
...
+const files = ['docusaurus.config.js', 'sidebars.js'];
+
+const start = async () => {
+  console.log(`Checking if the following files have been modified:
+${files.join('\n')}`);
+  const output = await getChanges();
+
+  const needsRegeneration = files.find((file) => {
+    return output.includes(`M ${file}`);
+  });
```
erickzhao, line 22: nit: `some` is a slightly better function here since we're using `needsRegeneration` as a boolean value.

```suggestion
  const needsRegeneration = files.some((file) => {
    return output.includes(`M ${file}`);
  });
```

i18n/readme.md
```diff
@@ -0,0 +1,6 @@
+# About en-US
+
+The locale `en-US` is used to keep the translations up to date when
+uploading the content to crowdin to avoid manually creating the
+different JSON files and dealing with inconsistencies.
+The contents are regenerated periodically.
```
erickzhao, line 6: Reading through this doc from the lens of someone without context, I have a few questions:

>  to avoid manually creating the different JSON files and dealing with inconsistencies.

Can we link the Docusaurus docs detailing what JSON files are required? Can we list an example of an inconsistency?

> The locale `en-US` is used to keep the translations up to date when uploading the **content** to crowdin [...] The **contents** are regenerated periodically.

Are these two referring to the same contents? What exactly are the contents? What process regenerates the contents, and how often is periodically?

molant: > Can we link the Docusaurus docs detailing what JSON files are required? Can we list an example of an inconsistency?

It depends on what plugins we have installed and they also could change if we update docusaurus so there is no "official" list. Does that make sense?

> Are these two referring to the same contents? What exactly are the contents? What process regenerates the contents, and how often is periodically?

The contents of the folder are these JSON files. I've added a pre-commit hook that checks if `docusuaurus.config.js` or `sidebars.js` have been modified and then forces the regeneration.

I'll make all of this clearer 😊

package.json
```diff
@@ -53,6 +55,7 @@
     "globby": "^11.0.3",
     "got": "^11.8.2",
     "gunzip-maybe": "^1.4.2",
+    "husky": "^6.0.0",
```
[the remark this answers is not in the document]
molant: I have no idea why 6 got installed. Just followed the instructions in their site 😓
Will take a look

i18n/readme.md
```diff
@@ -0,0 +1,20 @@
+# About en-US
+
+**TL;DR;** Do not manually modify the contents of `i18n/en-US`.
+
+Docusaurus translations for anything that is not markdown are stored
+in several JSON files. These files are generated from the
+contents of `docusaurus.config.js`, `sidebars.js`, and installed
+plugins.
```
erickzhao, line 8:
```suggestion
in several JSON files. These files are generated for each locale from
the contents of `docusaurus.config.js`, `sidebars.js`, and installed
plugins using the `yarn write-translations --locale [locale]` script.
```

i18n/readme.md
```diff
@@ -0,0 +1,20 @@
...
+Some of the files that are generated are `code.json`, `navbar.json`,
+`footer.json`, etc.
+
+Because those files only exist in a "translation" and not the "default"
+site, we create them in the `en-US` folder via
+`yarn write-translations --locale en-US` to uploaded them to crowdin via
+`electron/i18n` and then downloaded for the right locale.
```
erickzhao, line 15: suggestion: reworded this paragraph  to make things clearer and added relevant links

```suggestion
For the website's source language (`en-US`), these JSON files are **not**
used by Docusaurus. Rather, they serve as source files for our
[Crowdin](https://crowdin.com/project/electron) internationalization (i18n)
workflow. The [`electron/i18n`](https://github.com/electron/i18n) repository
periodically updates these source strings on Crowdin via
[GitHub Actions](https://github.com/electron/i18n/actions/workflows/schedule-update-source-content.yml).
```

.husky/.gitignore
```diff
@@ -0,0 +1 @@
+_
```
erickzhao, line 1: nit: this file can be removed as per changelog in https://github.com/typicode/husky/releases/tag/v7.0.0

erickzhao (commented):
Docs suggestions

.husky/.gitignore
molant: Removed!

erickzhao (approved):

PR #72, opened 2021-07-20
feat: activate blog plugin

### LINKED ISSUE #71: Enable blog
The blog needs to be migrated over from the old site to here as well. Docusaurus' has a blog plugin that we should be capable of leveraging.

The biggest problems when @erickzhao was investigating this previously were:

* Some markup not being allowed by Docusaurus (like `img` or `br` tags with no `/`)
* It does not support multiple authors (posts with multiple authors will be "signed" at the end)

molant:
Activate Docusaurus' blog plugin and make the previous posts available
in the new site.

Fix #71


ckerr (commented):
:+1:

I read the new code and setup, e.g. package.json and scripts/process-posts.js. If there are any changes in the blog posts themselves, I didn't read those.

ckerr (approved):

PR #78, opened 2021-07-21
feat: enable i18n

### LINKED ISSUE #64: Internazionalization

molant:
Integrate with Crowdin to download the latest translations and generate
the site for the supported locales.

Fix #64


molant:
This supersedes #64 and should be easier to review.

scripts/tasks/download-translations.js
```diff
@@ -0,0 +1,170 @@
...
+
+  for (let i = 0; i < counter; i++) {
+    build = await getBuild(buildId);
+    console.log(`Crowdin status: Project ${buildId} - ${build.status}`);
+
+    if (build.status === 'finished') {
+      break;
+    } else {
+      console.log(
+        `Crowdin status: Waiting ${interval} seconds (retry ${i}/${counter})`
```
ckerr, line 108: It doesn't look like there are many concurrent builds so this isn't a huge deal but it looks like https://support.crowdin.com/api/v2/#operation/api.projects.translations.builds.get might be more efficient / convenient than listProjectBuilds, since the former targets a single build id instead of returning all the builds.

For that matter, looks like you could just keep calling builds/{buildId}/download until it stops giving  you a 202. The nice thing is that payload returns both the status string for the console.log on line 102, but also a progress % that would be nice to have in the log too

scripts/tasks/download-translations.js
```diff
@@ -0,0 +1,170 @@
...
+/**
+ * Gets the download link for the latest available build
+ * @param {number} projectId
+ */
+const getLatestBuildLink = async (projectId) => {
+  const { data: builds } = await translationsApi.listProjectBuilds(projectId);
+
+  const {
+    data: { url },
+  } = await translationsApi.downloadTranslations(PROJECT_ID, builds[0].data.id);
```
ckerr, line 134: This assumes that the build at the front of the array is always the latest available. That may well be the case but I don't see that speced anywhere in the in crowdin docs. If you're sure this is correct, carry on :smile:

scripts/tasks/md-fixers.js
```diff
@@ -72,6 +72,19 @@ const fiddleTransformer = (line) => {
...
 
+/**
+ * Crowdin translations put markdown content right
+ * after HTML comments and thus breaking Docusaurus
+ * parse engine. We need to add a new EOL after `-->`
+ * is found.
+ */
+const newLineOnHTMLComment = (line) => {
+  if (line.includes('-->')) {
+    return line.replace('-->', '-->\n');
```
ckerr, line 83: Are there any edge cases that can bite us here -- 

- is there ever more than one html comment per line?

- will there ever already be a newline after the html comment? and if we make that a double newline, does that cause any problems with Docusaurus?

ckerr (commented):
I had a few questions inline -- some were optional changes, some were just questions I had to get familiar with the code -- but overall LGTM. No blockers.

scripts/pre-build.js
```diff
@@ -84,6 +82,39 @@ const start = async (source) => {
...
+  const locales = await downloadTranslations(I18N_FOLDER);
+
+  for (const locale of locales) {
+    const localeDocs = path.join(
+      I18N_FOLDER,
+      locale,
+      'docusaurus-plugin-content-docs',
+      'current'
+    );
+    const staticResources = ['fiddles', 'images'];
```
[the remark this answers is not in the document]
molant: Yes and no. Docusaurus is built in such a way that each locale is independent to enable exactly that.
The problem right now is that our i18n pipeline doesn't upload any of those assets to Crowding, so when I download the translations they are missing and I have to "manually" copy them from the source or otherwise they will be missing.

scripts/pre-build.js
```diff
@@ -84,6 +82,39 @@ const start = async (source) => {
...
+
+    console.log(`Copying static assets to ${locale}`);
+    for (const staticResource of staticResources) {
+      await fs.copy(
+        path.join(DOCS_FOLDER, staticResource),
+        path.join(localeDocs, 'latest', staticResource)
```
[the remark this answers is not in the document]
molant: Exactly!

A bit more context for anyone interested. The folder `current` is a Docusaurus convention to point to the latest version as you figured out (so `/docs/`). Because we don't want collision with the current documentation (www.electronjs.org/docs) and we want to have versioned docs sometime in the future, we append `/latest` so the "current" documentation is available via www.electronjs.org/docs/latest.

Older versions will be available via www.electronjs.org/docs/v12 (or a similar path yet TBD)

I haven't dig how the versioning of docs works exactly with docusaurus but its very likely it will be under another folder at the same level as `/current` like `/v12`.

scripts/tasks/download-translations.js
molant: For some reason I only get an array of 1 item no matter how many times I call it or kick a build. I will add a comment here explaining that.

scripts/tasks/md-fixers.js
molant: Good questions

> is there ever more than one html comment per line?

I've only seen an HTML comment in one of our markdown docs and unfortunately (and we cannot remove it for now). It was only one comment in one line

> will there ever already be a newline after the html comment? and if we make that a double newline, does that cause any problems with Docusaurus?

Double new lines are not a problem for Docusaurus. Crowdin already returns markdown that's slightly different than the original with extra lines.

scripts/tasks/download-translations.js
molant: Great suggestion!

scripts/pre-build.js
```diff
@@ -5,8 +5,13 @@
  * right places, and transform it to make it ready to
  * be used by docusaurus.
  */
+if (process.env.NODE_ENV !== 'production') {
+  require('dotenv').config();
+}
+
 const path = require('path');
 const { existsSync } = require('fs');
+const fs = require('fs-extra');
```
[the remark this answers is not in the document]
molant: I need to review when I'm deleting folders. I know that I liked it because it accepts globby patterns and then I could use `.gitignore` to clean all the files in the prebuild step but in this new PR it might not be needed.

molant:
The current build failures on Heroku are because of incorrect markdown being downloaded from Crowdin. I'll try to fix all of those before I'm on vacation 🏖

molant:
Build is failing now with the following message 😭

```
<--- Last few GCs --->
[4227:0x5f7aa80]  2083570 ms: Scavenge 2500.6 (2581.1) -> 2489.8 (2581.6) MB, 21.5 / 0.0 ms  (average mu = 0.315, current mu = 0.348) allocation failure 
[4227:0x5f7aa80]  2083755 ms: Scavenge 2503.7 (2581.6) -> 2493.1 (2582.6) MB, 24.5 / 0.0 ms  (average mu = 0.315, current mu = 0.348) allocation failure 
[4227:0x5f7aa80]  2087361 ms: Mark-sweep 2506.1 (2582.6) -> 2474.2 (2581.8) MB, 3465.3 / 0.1 ms  (average mu = 0.335, current mu = 0.354) allocation failure scavenge might not succeed
<--- JS stacktrace --->
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 0xa24ed0 node::Abort() [/tmp/build_36b48d1b/.heroku/node/bin/node]
 2: 0x966115 node::FatalError(char const*, char const*) [/tmp/build_36b48d1b/.heroku/node/bin/node]
 3: 0xb9acde v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [/tmp/build_36b48d1b/.heroku/node/bin/node]
 4: 0xb9b057 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [/tmp/build_36b48d1b/.heroku/node/bin/node]
 5: 0xd56ea5  [/tmp/build_36b48d1b/.heroku/node/bin/node]
 6: 0xd57a2f  [/tmp/build_36b48d1b/.heroku/node/bin/node]
 7: 0xd65abb v8::internal::Heap::CollectGarbage(v8::internal::AllocationSpace, v8::internal::GarbageCollectionReason, v8::GCCallbackFlags) [/tmp/build_36b48d1b/.heroku/node/bin/node]
 8: 0xd6967c v8::internal::Heap::AllocateRawWithRetryOrFailSlowPath(int, v8::internal::AllocationType, v8::internal::AllocationOrigin, v8::internal::AllocationAlignment) [/tmp/build_36b48d1b/.heroku/node/bin/node]
 9: 0xd37d2b v8::internal::Factory::NewFillerObject(int, bool, v8::internal::AllocationType, v8::internal::AllocationOrigin) [/tmp/build_36b48d1b/.heroku/node/bin/node]
10: 0x108035f v8::internal::Runtime_AllocateInYoungGeneration(int, unsigned long*, v8::internal::Isolate*) [/tmp/build_36b48d1b/.heroku/node/bin/node]
11: 0x1427079  [/tmp/build_36b48d1b/.heroku/node/bin/node]
Aborted
error Command failed with exit code 134.
```

molant:
We might need to build the languages one by one. There might be some leaking using the regular `build` command 🤔
Also the build takes forever (5-6min per language).

PR #81, opened 2021-08-10
fix: broken OpenJS Foundation logo

codebytere:
Fixes broken footer logo:

Before:
Screen Shot 2021-08-10 at 8 50 57 PM

After:
Screen Shot 2021-08-10 at 8 50 21 PM

Noticed this while looking at the new site which incidentally looks great :D


erickzhao (approved):
Thanks for this!!

PR #82, opened 2021-08-11
feat: link to external Releases homepage in navbar

### LINKED ISSUE #73: Releases section
We need to make the content of https://www.electronjs.org/releases/stable available in Docusaurus.

Some things to consider:
- I don't think URLs need to match (contrary to blog). They are not predictable at the moment because of pagination (one day the release x.y.z could be in page 1, the day after in page 2
- We should make it possible to link to each release individually
- There is also https://releases.electronjs.org, we should only have 1 place for releases and redirect to the right place where needed

erickzhao:
Closes #73 

Adds an external link to https://releases.electronjs.org

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -47,6 +47,11 @@ module.exports = {
         },
         { to: 'blog', label: 'Blog', position: 'left' },
         {
+          href: 'https://releases.electronjs.org',
+          label: 'Releases',
+          position: 'right',
+        },
+        {
           href: 'https://github.com/electron/electron',
           label: 'GitHub',
           position: 'right',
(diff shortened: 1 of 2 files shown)


molant:
LGTM, but I thought we wanted to consolidate and have less infrastructure/services and move that page here.

erickzhao:
> I thought we wanted to consolidate and have less infrastructure/services and move that page here.

After talking with @MarshallOfSound, there are a few dynamic features on https://releases.electronjs.org that would prevent all features from landing on the Docusaurus website:
* In-progress releases: https://releases.electronjs.org/
* Releases comparisons: https://releases.electronjs.org/release/compare/v13.0.0/v13.1.7

molant:
Should we look at adding the features from the electron page that aren't available into that one?
IIRC the release notes were more expanded and maybe something else?

erickzhao:
I think the full release notes are available with a click-through on releases.electronjs.org, although the UX could be improved to indicate this to users: https://releases.electronjs.org/release/v13.1.9

Issue #77, opened 2021-07-21
Homepage


molant:
We need to make the content of https://www.electronjs.org/ available in Docusaurus.

We should take this opportunity to rethink what we want to say in the homepage as it is not clear what the CTAs are for the user.

nabeelkhan10:
I am new to Open Source.
I want to contribute.
Can anyone help me assign an easy task?

PR #83, opened 2021-08-11
feat: add home page

erickzhao:
Ref #77 

Adds a simple hero and SCSS support. A later PR should decide what marketing copy we should put onto the homepage, but this should be a marked improvement over the blank page from before.


ckerr (approved):
Makes sense to me

PR #85, opened 2021-08-13
refactor: use CSS modules

erickzhao:
Refactors the custom styling added in #83 to use [CSS modules](https://github.com/css-modules/css-modules) rather than a global stylesheet with BEM naming.

A couple advantages to this approach:
* Less verbose style rules.
* No possibility of rule conflicts between modules.
* Easy distinguishing between builtin Docusaurus classes and custom CSS.

Some notes:
* Styles are written using `camelCase` so that they can be called easily through the JSX.
* Some global styles remain from before the last PR.
* Other change: removed numbered classes used to animate the hero image (e.g. `.hero-image__app-N`). Instead use `:nth-child()` selectors for the animations.


src/pages/index.module.scss
```diff
@@ -0,0 +1,96 @@
+.heroElectron {
+    color: var(--electron-color-light);
+    background: var(--electron-color-dark);
+}
+
+@keyframes hero-animation {
```
molant, line 6: If we have to use camel casing for the class names should we use it also everywhere in the CSS?

molant (approved):
LGTM, just have one non-blocking question

src/pages/index.module.scss
erickzhao: That's a good point. I'll amend this to keep it uniform :)

erickzhao: fb85d488820c3e3bfbe0cc3680237abd8e961543

PR #102, opened 2021-09-15
feat: add badge markup for API labels

### LINKED ISSUE #101: Transform platform notices in API docs into badges
In https://github.com/electron/electronjs.org, we parse through the raw API docs to add badges to the headings. These badges add a bit of extra info about each API (platform, deprecation, and experimental status).

https://github.com/electron/electronjs.org/blob/224c921864f99179944b4bf11f57d242462a2972/scripts/docs-api-labels.js#L26-L48

image

These should also be added to the new website. Probably during the site generation step by manually using the Infima badge styles: https://infima.dev/docs/components/badge/

erickzhao:
Closes #101 

image

### DIFF
--- a/src/transformers/api-labels.js
+++ b/src/transformers/api-labels.js
@@ -55,7 +55,7 @@ async function transformer(tree) {
       } else if (tag === EXPERIMENTAL) {
         node.data = {hProperties: {className: ["badge badge--warning"]}}
       } else if (tag === READONLY) {
-        node.data = {hProperties: {className: ["badge badge--secondary"]}}
+        node.data = {hProperties: {className: ["badge badge--info"]}}
       }
     }
   }
--- a/src/transformers/api-labels.js
+++ b/src/transformers/api-labels.js
@@ -1,0 +1,39 @@
+//@ts-check
+
+const visitParents = require("unist-util-visit-parents");
+
+module.exports = function attacher() {
+  return transformer;
+};
+
+const PLATFORMS = ['macOS', 'Windows', 'Linux'];
+const DEPRECATED = 'Deprecated';
+const EXPERIMENTAL = 'Experimental';
+const READONLY = 'Readonly';
+
+/**
+ *
+ * @param {import("unist").Parent} tree
+ */
+async function transformer(tree) {
+  visitParents(tree, 'emphasis', visitor);
+}
+
+  /**
+   *
+   * @param {import("unist").Node} node
+   */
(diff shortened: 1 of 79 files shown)


src/transformers/api-labels.js
```diff
@@ -0,0 +1,39 @@
...
+    if (Array.isArray(node.children) && node.children.length === 1) {
+      const tag = node.children[0].value;
+      if (PLATFORMS.includes(tag)) {
+        node.data = {hProperties: {className: ["badge badge--primary"]}}
+      } else if (tag === DEPRECATED) {
+        node.data = {hProperties: {className: ["badge badge--danger"]}}
+      } else if (tag === EXPERIMENTAL) {
+        node.data = {hProperties: {className: ["badge badge--warning"]}}
+      } else if (tag === READONLY) {
+        node.data = {hProperties: {className: ["badge badge--secondary"]}}
```
MarshallOfSound, line 36:
```suggestion
        node.data = {hProperties: {className: ["badge badge--info"]}}
```

I think `secondary` colors are too close to the code background color, info is distinct enough 👍 

E.g. this is hard to separate visually
image

src/transformers/api-labels.js
```diff
@@ -0,0 +1,39 @@
...
+async function transformer(tree) {
+  visitParents(tree, 'emphasis', visitor);
+}
+
+  /**
+   *
+   * @param {import("unist").Node} node
+   */
+  function visitor(node) {
+    if (Array.isArray(node.children) && node.children.length === 1) {
```
molant, line 27: Do we want to filter by node type as well? Will this also transform into a badge a paragraph that only has `Windows`?

I think it would be good for future maintainers to have a small description on where do we expect this transformer to be executed, what it is doing and why.

Maybe something like:

> This transformer should be executed only in the API pages and transforms XYZ into badges to provide visual cues for the platform and status

erickzhao: This will only execute when the tree node is:
* an `_emphasis_` node
* with exactly 1 child
* that child's value is specifically one of the strings that we're matching for

e.g. `_macOS_` or `_Windows_`

I think that's precise enough

erickzhao: Added documentation in fb2fa3e

molant: I completely missed the `_emphasis_` at the top, my bad 😓

molant (approved):

PR #108, opened 2021-09-23
feat: clean up API headings in TOC

### LINKED ISSUE #98: Right sidebar formatting makes really hard to quickly glance for methods
New documentation makes right-sidebar formatting unusable:
1. Unnecessary `code` style for method names. 
2. Arguments added to the list of methods. JS does not support method overloading - this is not needed.
3. They don't fit on small screens.

Suggestins:
- Remove `code formatting` use default link style.
- Remove method arguments.
- Remove common prefix, e.g. `webFrame.setZoomFactor()` vs `setZoomFactor`.
Here is an example of new doc and old doc:
image

erickzhao:
Fixes #98.

Because Docusaurus doesn't expose any modification for the TOC content, I went ahead and swizzled the `@theme/TOC` component to clean the content within a wrapper component.

Note that this is IMO a low-risk change but will break if `TOCItem` from `@docusaurus/types` ends up changing:

https://github.com/facebook/docusaurus/blob/a52d49921817a159e291d1bbf05e8beaa7b4b8fd/packages/docusaurus-types/src/index.d.ts#L420-L424

Screenshot:
image


src/theme/TOC/styles.module.css
```diff
@@ -0,0 +1,23 @@
+/**
```
molant, line 1: Can you add a notice saying the package where this file is coming from?
It might become handy if we have style problems when updating.

molant (approved):
LGTM!
I think it just needs a comment and then you can merge directly.

src/theme/TOC/styles.module.css
erickzhao: Sure :)

PR #112, opened 2021-10-07
feat: Crowdin integration

### LINKED ISSUE #64: Internazionalization

molant:
Integrates with Crowdin to enable i18n of the site. This PR changes the
source of truth for Crowdin and moves it to this repository instead of
relying on `electron/i18n`.

Additionally it uses Crowdin's CLI to do the upload/download of assets
making it more reliable.

The website is built on locale at a time via `yarn i18n:build`.
Otherwise the process crashed with an out of memory error. The regular
`yarn build` command still compiles the `en` locale.

Because we cannot get notifications when there are new translations
avaiable, there is a GitHub workflow (`update-i18n-deploy.yml`) that
downloads the content every few minutes, builds, and deploy. To speed up
this process, the previous generated assets are download. In local tests
this reduces the build times from 250s to 40s so the whole process
should take about 5 minutes.

The previous generated content is stored in Azure Storage. Because this
is a static website it makes more sense than having a dyno and will make
it easier to:
- deploy multiple locales at the same time if we still need to speed up
  the process
- have versioned docs because we just need to "take a snapshot" and
  publish to a different folder

The current live site is still not using this storage but will soon-ish.

Fix #64

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -56,10 +56,11 @@ module.exports = {
           label: 'Releases',
           position: 'right',
         },
-        {
-          type: 'localeDropdown',
-          position: 'right',
-        },
+        // TODO: uncomment once the production site is pointing to Azure Storage
+        // {
+        //   type: 'localeDropdown',
+        //   position: 'right',
+        // },
         {
           href: 'https://github.com/electron/electron',
           label: 'GitHub',


molant:
@erickzhao the markdown changes add an automatic description to a bunch of documents that didn't have one and I think they can be ignored. I split the PR into 2 commits so it's easier to review.

PR #115, opened 2021-10-13
fix: respect system color theme (dark/light)

### LINKED ISSUE #114: Doesn't switch by default to dark mode according to System theme
## Describe the Bug

The electronjs.org website doesn't automatically open in Dark mode if my system is in Dark Mode when opening the website for the first time. 
But the website **does** switch to dark mode if my system changes to dark mode while the website is open though which is the right behaviour.

## A Demo

The following demo uses Incognito windows to simulate first-time visit behaviour.

https://user-images.githubusercontent.com/55079486/137184260-bcfd5212-54b8-4111-b0f2-e7b7108be0c2.mp4

## Possible Solution

I went through the Docusaurus documentation and saw that adding a parameter ```respectPrefersColorScheme: true``` made the website appear by default in Dark mode

Reference: https://docusaurus.io/docs/api/themes/configuration

AJAYK-01:
## What does this PR do?

Fixes #114 

## Description of the change

Adds ```respectPrefersColorScheme: true``` property in [docusaurus.config.js](https://github.com/electron/electronjs.org-new/blob/main/docusaurus.config.js) which sets the website to use Light or Dark theme according to the System theme on the first load of the website.

Reference: https://docusaurus.io/docs/api/themes/configuration

## Demo

https://user-images.githubusercontent.com/55079486/137192868-38cabeb2-f334-4c01-a8f9-1aaa1d320d28.mp4

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -18,6 +18,10 @@ module.exports = {
     locales: ['en', 'de', 'es', 'fr', 'ja', 'pt', 'ru', 'zh'],
   },
   themeConfig: {
+    colorMode: {
+      //Default to light or dark depending on system theme.
+      respectPrefersColorScheme: true,
+    },
     navbar: {
       title: 'Electron',
       logo: {


molant:
Should be up and running!
https://github.com/electron/electronjs.org-new/runs/3886687169?check_suite_focus=true

PR #119, opened 2021-10-14
fix: navbar link to homepage

groundwater:
I humbly request that we add this for now, so that clicking the Electron icon takes you to the real homepage, and not the SPA placeholder.


Selena20191:
> I humbly request that we add this for now, so that clicking the Electron icon takes you to the real homepage, and not the SPA placeholder.

PR #120, opened 2021-10-15
feat: versioned docs

### LINKED ISSUE #118: Versioned docs
The website only shows the latest stable documentation, but the reality is that developers are not always using the latest and brightest version. While it is possible to go to `electron/electron` and find the docs for a particular branch, we could make the process easier.

Docusaurus allows you to have [versioned docs](https://docusaurus.io/docs/versioning). That said, the process doesn't really fit our needs:

* We have to keep old doc versions in our repo and this will increase the build time and probably crash things as it happened with the localization
* I'm not even sure how that will work with localized content
* Even though it should be possible to change the version from `Next` to something else, I was not capable. Maybe it was because I only had one version
* We do want to "build and forget", meaning that the website should be a snapshot of whatever it was

My proposal is to leverage the fact we are using a storage solution now to publish the website and:

- Add a new [`dropdown` item to `docusaurus.config.js`](https://docusaurus.io/docs/api/themes/configuration#navbar-dropdown) with the versions we want developers to have quick access. Even if we publish more versions and are available, I think we should only list the supported ones here.
- When there is a major release, before switching we build the latest one again (localized content and everything) but instead of publishing to `latest` we publish to `vXX`. This might require us to change `docusaurus.config.js` and `sidebars.js`.
- Publish to the storage
- Update `docusaurus.config.js` dropdown with the new version pointing to `https://www.electronjs.org/vXX`
- Continue the regular docs process

### How do we handle minor release doc updates (e.g. 14.0.1 to 14.1.0) for a version that is not the latest one?

We could generate a build each time, publish under vX.Y.Z. In `docusaurus.config.js` we only list the latest version of each major so the list is not too long, but the content will still be available.

The biggest problem is with the localized content. Crowdin only handles the latest stable content and not all content is translated. On top of that, we do not keep the translations in this repo...

A few options (from most complicated to less complicated) to handle this scenario:

1. We start doing versioning on Crowdin. Not very straightforward. Also, the translations take some time so we will have also to find a way to get changes to previous versions and publish those as well and thus complicating the whole process
1. Use the latest version in Crowdin. While it should be mostly OK (at least when there are bumps in minors), there could be breaking changes in the APIs and thus report stuff that is not accurate). 
1. We only build the EN locale for previous versions

My vote is with the last option, and if developers really demand versioned docs for previous versions, then we look at implement it (start small).

molant:
The way versioned documents are supported on the site differs a bit from
Docusaurus' documentation:

- Each doc change to a branch is build and publish separately.
- The links to the versions are `_blank` to bypass `react-router` even
  though the content is in the same server. The reason is that because
  it is built at different steps it is not "found".

This is achieve thanks to `electron-website-updater` sending a
`repository_dispatch` event with 2 different types of actions for each
`push` event that happens in `electron/electron` that contains doc
changes. The 2 types of events are:

* `doc_changes_branches`: Updates to documentation branches (does not
  matter if they are the latest or not).
* `doc_changes`: Updates to the latest stable branch.

The tasks to perform are almost identical and are split in two different
GitHub actions for each event:

1. Download the markdown for the SHA, check if there are new version
   branches and update `versions-info.json`, update the SHA, and publish
   those changes to Git.
   This is done in `update-docs-XXX.yml` when we receive the event.
2. Build the content for that branch and publish in the right place.
   This is done in `push-XXX.yml` when there is a push to `main` or a
   `vXX` branch.

The deployment is done to the storage service. The difference is that a
branch will only push `assets` and `docs` while `main` will publish
everything.

Additionally, Crowdin gets updated every time there is a change in
`main` to make sure the latest content is always uploaded.

Fix #118

### DIFF
--- a/.github/workflows/push-main.yml
+++ b/.github/workflows/push-main.yml
@@ -11,14 +11,17 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-
       - name: Use Node.js 14
         uses: actions/setup-node@v2
         with:
           node-version: 14
-
       - name: Install dependencies
         uses: bahmutov/npm-install@HEAD
+
+      - name: Upload sources to Crowdin
+        run: 'yarn i18n:upload'
+        env:
+          CROWDIN_PERSONAL_TOKEN: ${{ secrets.CROWDIN_PERSONAL_TOKEN }}
 
       - name: Test
         run: yarn test
--- a/.github/workflows/push-others.yml
+++ b/.github/workflows/push-others.yml
@@ -1,0 +1,37 @@
+name: Push on version branch
+
+on:
+  push:
+    branches:
+      - 'v**'
+      - '!versioned-docs'
+
+jobs:
+  tests:
+    name: Push updates to previous versions
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set environment variables
+        run: |
+          echo "GIT_BRANCH=${GITHUB_REF##*/}" >> $GITHUB_ENV
+
+      - uses: actions/checkout@v2
+      - name: Use Node.js 14
+        uses: actions/setup-node@v2
+        with:
+          node-version: 14
+      - name: Install dependencies
+        uses: bahmutov/npm-install@HEAD
+
+      - name: 'Build branch version (EN)'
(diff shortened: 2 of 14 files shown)


molant:
Need to figure out what happens with the sidebars when we are publishing a previous version and there are "missing documents" 😓

PR #126, opened 2021-10-20
style: exclude headings from inline code color

erickzhao:
| Before | After |
|--------|-------|
|image | image|

Too much red in the initial implementation!


erickzhao:
I think it helps to break the wall of navy blue here by adding a secondary colour. I've made the colour code darker on light theme.

cd4e344

PR #136, opened 2021-10-29
feat: add governance page

erickzhao:


src/pages/governance/index.jsx
```diff
@@ -0,0 +1,88 @@
...
+      </div>
+      <div className="card__body">
+        <p className={styles.cardSubtitle}>Members</p>
+        <div className={clsx(styles.cardList, hasExpand && (expanded ? styles.expanded : styles.unexpanded))}>
+          <Member user={group.chair} isChair/>
+          { group.members.map(user => (<Member key={user} user={user}/>)) }
+        </div>
+      </div>
+      {/** Only show this button if needed. Apply visibility:hidden so it takes up the same height. */}
+      <button className={clsx("button", styles.buttonSubtle, !hasExpand && styles.hidden)}
```
molant, line 62: ▲ or ▼ are not really accessible.

src/pages/governance/index.jsx
```diff
@@ -0,0 +1,88 @@
...
+  )
+};
+
+const Member = ({ user, isChair }) => {
+  return (
+    <div className="avatar" style={{margin: '1rem'}}>
+      <img
+        className="avatar__photo avatar__photo--sm"
+        src={`https://github.com/${user}.png`}
+      />
```
molant, line 81:
```suggestion
     alt="" />
```

They are decorative images so we can have no alternate text (which is better than not having anything)

src/css/custom.scss
```diff
@@ -17,6 +17,7 @@
   --ifm-color-primary-light: #00aeff;
   --ifm-color-primary-lighter: #4cc8ff;
   --ifm-color-primary-lightest: #97e1ff;
+  --ifm-color-secondary-darkest: #808080;
```
molant, line 20: We need something with more contrast. This doesn't pass WCAG 2 AA (4.5:1) . The contrasts are:
- 3.88 when used in the cards (#242526), 
- 4.45 when outside the cards (just below "Electron Governance")

src/pages/governance/index.jsx
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * See "governance" script for _data.json generation
+ */
+import data from './_data.json';
+
+export default function GovernancePage() {
+  return (
+    <Layout title="Governance">
+      <div className="container margin-vert--xl">
```
molant, line 15:
```suggestion
      <main className="container margin-vert--xl">
```

We are missing a main landmark and I believe there is a "hidden link" to jump to the main content in docusaurus as well.

molant (changes requested):
There are just some accessibility issues that hopefully should be easy to fix.

Also what do you think of the following:
image

It will fix another a11y issue where we are using `GitHub` to point to different places, avoid the repetition in all cards, and give us a bit more real state.

scripts/governance.js
```diff
@@ -0,0 +1,136 @@
...
+ * Fetch the contents of each `electron/governance` working group
+ * README.md file, parse the contents using `remark`, and write the
+ * output array to `_data.json`.
+ */
+async function main() {
+  console.log('Fetching governance data...')
+  const data = JSON.stringify(await fetchGovernanceData());
+  const targetPath = path.join(__dirname, '..', 'src', 'pages', 'governance', '_data.json');
+  console.log('Writing to disk...');
+  fs.writeFileSync(targetPath, data);
```
molant, line 24:
```suggestion
  await fs.writeFile(targetPath, data);
```

scripts/governance.js
```diff
@@ -0,0 +1,136 @@
...
+  const headers = {};
+  if (process.env.GITHUB_TOKEN) {
+    headers.Authorization = `Bearer ${process.env.GITHUB_TOKEN}`
+  }
+  // @ts-ignore: export error?
+  const res = await got(`https://api.github.com/repos/electron/governance/contents/wg-${workingGroup}/README.md`,
+    { headers }
+  );
+  const { content } = JSON.parse(res.body);
+  return Buffer.from(content, 'base64').toString();
```
molant, line 124: Can you explain why we are doing the following? Is this something related to remark or something like that?

String (res.body) --> JSON object --> Buffer --> String

molant (changes requested):
Reviewed now the scripts to get the data and have a couple questions.

scripts/governance.js
erickzhao: Simplified a bit and added comments in a761429

src/css/custom.scss
erickzhao: Digging into the Infima code (undocumented), I found some variables that work. 37afe40

src/pages/governance/governance.module.scss
```diff
@@ -0,0 +1,123 @@
...
+  color: var(--ifm-color-white);
+  line-height: var(--ifm-avatar-photo-size-sm);
+  // one-off colors
+  background: linear-gradient(to top, #dd5e89, #f7bb97);
+}
+
+.cardDescription {
+  display: block;
+  font-size: 0.9em;
+  min-height: 8em;
```
molant, line 69:
```suggestion
  min-height: 2em;
```

I think we can reduce the height and still have it look OK. What do you think?

| 8em | 2em |
| -- | -- |
| Card with description min-height 8em | Card with description min-height 2em |

erickzhao: The 8em height is to accommodate the longer copy in the Security WG README. It's about twice as long as the other ones, but I wanted all cards to remain the same height.

PR #145, opened 2021-11-10
docs(draft): add IPC tutorial

erickzhao:


docs/latest/tutorial/ipc.md
```diff
@@ -0,0 +1,554 @@
...
+});
+```
+
+At this point, your demo should be fully functional. Try using the input field and see what happens
+to your BrowserWindow title!
+
+## Pattern 2: Renderer to main (two-way)
+
+A common application for two-way IPC is calling a main process module from your renderer process
+code and waiting for a result. This can be done easily by using [`ipcRenderer.invoke`] paired with
```
Kilian, line 171: scratch "easily"?

docs/latest/tutorial/ipc.md
```diff
@@ -0,0 +1,554 @@
...
+```
+
+[`ipcRenderer.invoke`]: ../api/ipc-renderer.md#ipcrendererinvokechannel-args
+[`ipcMain.handle`]: ../api/ipc-main.md#ipcmainhandlechannel-listener
+
+### 1. Listen for events with `ipcMain.handle`
+
+In the main process, we'll be creating a `handleFileOpen()` function that calls
+`dialog.showOpenDialog` and returns the value of the file path selected by the user. This function
+is used as a callback whenever an `ipcRender.invoke` message is sent through the `dialog:openFile`
```
Kilian, line 191: Might be nice to add a small comment at this point how `foo:bar` is a pattern used to "namespace" channel names, but that the `:` itself is not special and doesn't change the behavior.

docs/latest/tutorial/ipc.md
```diff
@@ -0,0 +1,554 @@
...
+ipcRenderer.on('asynchronous-reply', (_event, arg) => {
+  console.log(arg) // prints "pong" in the DevTools console
+})
+ipcRenderer.send('asynchronous-message', 'ping')
+```
+
+```javascript title='main.js (Main Process)'
+ipcMain.on('asynchronous-message', (event, arg) => {
+  console.log(arg) // prints "ping" in the Node console
+  event.reply('asynchronous-reply', 'pong')
```
Kilian, line 326: Add a short note on how `reply` works? (even if it's "the same as send, but...")

docs/latest/tutorial/ipc.md
```diff
@@ -0,0 +1,554 @@
...
+Before proceeding to implementation details, you should be familiar with the idea of using a
+[preload script] to import Node.js and Electron modules in a context-isolated renderer process.
+
+* For a full overview of Electron's process model, you can read the [process model docs].
+* For a primer into exposing APIs from your preload script using the `contextBridge` module, check
+out the [context isolation tutorial].
+
+[preload script]: process-model.md#preload-scripts
+[process model docs]: process-model.md
+[context isolation tutorial]: context-isolation.md
```
molant, line 39: Once the guide is OK can we move all the label links at the bottom of the doc?
I've had some issues when refactoring the docs in the tutorials and updating to new references.

docs/latest/tutorial/ipc.md
molant:
```suggestion
code and waiting for a result. This can be done by using [`ipcRenderer.invoke`] paired with
```

docs/latest/tutorial/ipc.md
```diff
@@ -0,0 +1,554 @@
...
+  const filePath = await window.electronAPI.openFile()
+  filePathElement.innerText = filePath
+})
+```
+
+In the above snippet, we listen for clicks on the `#btn` button, and call our
+`window.electronAPI.openFile()` API to activate the native Open File dialog. We then display the
+selected file path in the `#filePath` element.
+
+### Note: alternative approaches
```
molant, line 295: From what I'm reading none of the other approaches are recommended. Should we really document them?
IPC is already a hard topic, if we add multiple ways of doing things maybe we risk confusing them?

docs/latest/tutorial/ipc.md
```diff
@@ -0,0 +1,554 @@
...
+The structure of this code is very similar to the `invoke` model, but we recommend
+**avoiding this API** for performance reasons. Its synchronous nature means that it'll block the
+renderer process until a reply is received.
+
+## Pattern 3: Main to renderer
+
+When sending messages from the main process to the renderer process, there is no direct equivalent
+to `ipcRenderer.send` in the `ipcMain` module. Messages need to be sent to individual [WebContents]
+instances because you can have more than one renderer in your Electron app and you might have web
+embeds running in any given renderer.
```
molant, line 367: We are talking about main process, renderer process, WebContents, modules, and saying there is no direct equivalent (although it is just in a module). This can be a bit scary for someone not familiar. 
What do you think of something around the following (there's probably a better way to phrase it, this is just to give an idea):

```suggestion
In an Electron application there is only one main process, but there can be multiple renderer processes.
Messages are sent to individual renderer process via its [WebContents] instance and the `send` method, making the
call similar to `mainWindow.webContents.send`
```

docs/latest/tutorial/ipc.md
```diff
@@ -0,0 +1,554 @@
...
+[breaking changes]: ../breaking-changes.md
+
+### Renderer-to-renderer communication
+
+There's no direct way to send messages between renderer processes in Electron using the `ipcMain`
+and `ipcRenderer` modules. To achieve this, you have two options:
+
+* Use the main process as a message broker between renderers. This would involve sending a message
+from one renderer to the main process, which would forward the message to the other renderer.
+* Pass a [MessagePort] from the main process to both renderers.
```
molant, line 552: Maybe mention that MessagePort should be a more efficient because it avoids the jump through main?

docs/latest/tutorial/ipc.md
```diff
@@ -0,0 +1,554 @@
...
+
+> In particular, DOM objects (e.g. `Element`, `Location` and `DOMMatrix`), Node.js objects
+> backed by C++ classes (e.g. `process.env`, some members of `Stream`), and Electron objects
+> backed by C++ classes (e.g. `WebContents`, `BrowserWindow` and `WebFrame`) are not serializable
+> with Structured Clone.
+
+[sca]: https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm
+[breaking changes]: ../breaking-changes.md
+
+### Renderer-to-renderer communication
```
molant, line 545: What do you think of moving this a level up as a pattern?
At the end what developers want to know is how to communicate different process regardless of what module is being used.

docs/latest/tutorial/ipc.md
```diff
@@ -0,0 +1,554 @@
...
+//...
+ipcMain.on('counter-value', (_event, value) => {
+  console.log(value) // will print value to Node console
+})
+//...
+```
+
+## Limitations of IPC
+
+### Object serialization
```
molant, line 529: I would remove the title `## Limitations of IPC` and move this up.

docs/latest/tutorial/ipc.md
```diff
@@ -0,0 +1,554 @@
...
+For performance reasons, Electron's IPC implementation uses the HTML standard
+[Structured Clone Algorithm][sca] to serialize objects passed between processes, meaning that
+only certain types of files can be passed through IPC channels.
+
+This was a breaking change introduced in Electron 8. From our [Breaking Changes] doc:
+
+> In particular, DOM objects (e.g. `Element`, `Location` and `DOMMatrix`), Node.js objects
+> backed by C++ classes (e.g. `process.env`, some members of `Stream`), and Electron objects
+> backed by C++ classes (e.g. `WebContents`, `BrowserWindow` and `WebFrame`) are not serializable
+> with Structured Clone.
```
molant, line 540: Could these type of objects be sent before? I thought before the messages were JSON strings so I wouldn't expect many of those be also serializable. Am I wrong?

I'd like this to read this under a more possitive lens. What about doing something like:

-  We used JSON string earlier so we switched to SCA that bring a perf boost. That said not everything can be sent: DOM objects, Node.js objects...
- Remove the breaking change mention. Electron 8 is way in the past

erickzhao: > Could these type of objects be sent before? I thought before the messages were JSON strings so I wouldn't expect many of those be also serializable. Am I wrong?

From the breaking changes doc, it seems like these were serializable up to Electron 9:
> Sending any objects that aren't native JS types, such as DOM objects (e.g. Element, Location, DOMMatrix), Node.js objects (e.g. process.env, Stream), or Electron objects (e.g. WebContents, BrowserWindow, WebFrame) is deprecated. In Electron 8, these objects will be serialized as before with a DeprecationWarning message, but starting in Electron 9, sending these kinds of objects will throw a 'could not be cloned' error.

docs/latest/tutorial/ipc.md
erickzhao: I wouldn't mind. As an aside, it'd probably be good to put more detail into this section in the future.

docs/latest/tutorial/ipc.md
erickzhao: > move this up.

As in `## Object serialization`, or change its position in the doc (or both)?

docs/latest/tutorial/ipc.md
erickzhao: I think it's important to note, especially since the APIs aren't deprecated and people might wonder about the difference between all the different methods.

molant: From what I've seen developers have a hard time getting their head around IPC (with reason!). Anything we can do to make it easier for them the better and I think that mentioning only the recommended approaches is better to set them up for success.
If you feel really strong about documenting all approaches, maybe we can move this down the doc and rename it to something like "Non-recommended alternative approaches"?

docs/latest/tutorial/ipc.md
molant: As in `## Object serialization` 😊

docs/latest/tutorial/ipc.md
molant: Thanks for digging this up.
I still think we should remove the quote from the breaking changes and just mention what type of objects aren't cloneable.

docs/latest/tutorial/ipc.md
erickzhao: I wouldn't move it to a different spot because these examples apply specifically for this pattern, but I think it's a good idea to change the wording here to dissuade people from taking these approaches.

PR #166, opened 2022-01-20
chore: update to DocSearch V3 index

erickzhao:


docusaurus.config.js
```diff
@@ -160,7 +160,7 @@ module.exports = {
       copyright: `Copyright © ${new Date().getFullYear()} OpenJS Foundation and Electron contributors.`,
     },
     algolia: {
-      apiKey: 'c9e8f898b3b32afe40f0a96637e7ea85',
+      apiKey: 'fdc2cf6080e499639d7e6b0278851ed4',
```
shortcuts, line 163: You should also add the `appId` we provided in the email.

docusaurus.config.js
```diff
@@ -160,7 +160,7 @@ module.exports = {
...
       indexName: 'electronjs',
```
shortcuts, line 164: Isn't this repo for https://beta.electronjs.org/? We also have a crawler for it

docusaurus.config.js
shortcuts: So it is either

index name: `electronjs`
```suggestion
      appId: 'MG3SRMK3K0',
      apiKey: 'c9e8f898b3b32afe40f0a96637e7ea85',
```

or

index name: `beta-electronjs`
```suggestion
      appId: '4PUIL4O0HL',
      apiKey: '681405e5bb1f0ab10cff4f14b94217b7',
```

erickzhao: For the first one, would it be `fdc2cf6080e499639d7e6b0278851ed4` for the `apiKey`?

shortcuts: Indeed you are right sorry for the confusion

PR #174, opened 2022-02-11
chore: update target browsers

erickzhao:


molant:
@erickzhao is there something else that needs to get changed other than `yarn.lock`? I don't understand how updating that file "updates target browsers"

erickzhao:
I believe Docusaurus uses `browserslist` to determine the transpilation target. https://docusaurus.io/docs/browser-support#default-values

Since we don't explicitly define browser versions in our `package.json`, browserslist relies on `caniuse` versions, which are the transient dependencies by running `npx browserslist@latest --update-db`.

PR #175, opened 2022-02-13
build: Parallelize building locale

MikuroXina:
Building a locale seems taking a long time in the workflow. So I tried to parallelize the process building locale files. But I'm not sure to resolve the problem by this.

If this change is not needed, I don't mind to close this PR without a notice.

### DIFF
--- a/scripts/i18n-build.js
+++ b/scripts/i18n-build.js
@@ -6,18 +6,28 @@ const {
   i18n: { locales, defaultLocale },
 } = require('../docusaurus.config');
 
+/**
+ *
+ * @param {string} [locale]
+ */
 const updateConfig = async (locale) => {
   const baseUrl = locale !== defaultLocale ? `/${locale}/` : '/';
   const configPath = join(__dirname, '../docusaurus.config.js');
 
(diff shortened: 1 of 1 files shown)


molant:
Thanks @MikuroXina,

Have you tried running this locally and comparing to the sequential times?
If it's faster, only comment would be to figure out a way to print the right locale if it fails.

MikuroXina:
Hmm, it looks got worse. Should I close this?

Benchmarked on my laptop with `hyperfine`:

Before:
  Time (mean ± σ):     210.119 s ± 132.326 s    [User: 318.679 s, System: 27.513 s]
  Range (min … max):   166.173 s … 586.647 s    10 runs

After:
  Time (mean ± σ):     270.450 s ± 133.260 s    [User: 1110.374 s, System: 74.988 s]
  Range (min … max):   99.203 s … 437.842 s    10 runs

molant:
@MikuroXina I believe the hardware where our CI runs is going to be worse than your system so we should probably not parallelize anything. It was worth a shot though, thanks!

PR #177, opened 2022-02-15
feat: custom sidebar metadata

erickzhao:
Electron has a lot of platform-specific APIs, and guides to go along with them.
A lot of these guides have long-winded suffixes indicating which platforms they
apply to (e.g. `(Windows and macOS)`). This information is more scannable with
visual imagery instead of text.

We can attach metadata onto `sidebars.js` items via the `customProps` object,
but the default `DocSidebarItem` component has no way to act on the props being
passed in, so we must Swizzle.

See: https://docusaurus.io/docs/sidebar#passing-custom-props

The `customProps` object will accept a `tags` property of type
`Array<'mac'|'windows'|'linux'>`. These will be rendered as SVG icons that are
stored as static assets in `static/assets/img/`.

```js
// sidebars.js
{
  type: 'doc',
  id: 'doc1',
  customProps: {
    tags: ['mac', 'windows', 'linux']
  },
};
```


src/theme/DocSidebarItem/TagContent.jsx
```diff
@@ -7,11 +7,11 @@ import WindowsIcon from '@site/static/assets/img/platform-windows.svg';
 export default function TagContent({ platform }) {
   switch (platform) {
     case 'windows':
-      return <WindowsIcon role="img"/>;
+      return <WindowsIcon role="img" title="Available on Windows"/>;
```
molant, line 10: > Available on ...

or

> Works on...

molant (approved):
LGTM!

molant:
Copy pasting the images in slack for posterity and easier review:

| Before | After |
| --- | --- |
| ToC Before | ToC After |

PR #180, opened 2022-02-23
fix: admonitions from crowdin

molant:
Crowdin sometimes breaks the admonitions closing tags (`:::`)
and puts them in the same line instead of a new one. This breaks
the build process and times out our i18n deployment.

### DIFF
--- a/scripts/tasks/md-fixers.js
+++ b/scripts/tasks/md-fixers.js
@@ -121,6 +121,7 @@ const transform = (doc) => {
     fiddleTransformer,
     newLineOnHTMLComment,
     newLineOnAdmonition,
+    fixAdmonitions,
   ];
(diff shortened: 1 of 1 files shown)


molant:
@erickzhao I've done this directly from GitHub and haven't had the time to run it locally (don't have the environment setup in my only machine and electron does not have codespaces 😭).

erickzhao (changes requested):
Duplicate of adf029c3cbd572c2e191775ff3bd3a775e8b8868 ?

molant:
Then it is not enough because I'm still receiving building errors when the action is run 😭

erickzhao:
Going to investigate further when I have the time. Closing this PR for now. :)

PR #183, opened 2022-03-09
feat: added shorthand to preset

Biki-das:
The docusaurus.config file has been supporting `module shorthands` , and electrons docusaurus config is following the same, the preset was not using it ,so i made it shorthand


erickzhao (changes requested):
I personally don't like this change since it makes the preset name less explicit and we don't really improve much by shortening this string.

molant:
I think the same as @erickzhao.
Thanks for proposing though!

PR #184, opened 2022-03-09
feat: added configuration for author and time updated

Biki-das:
I have added a small configuration which is available to docusaurus , it shows the author name who last updated the doc and the time

Notifications _ Electron - Google Chrome 3_9_2022 1_05_51 PM

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -173,6 +173,8 @@ module.exports = {
         docs: {
           sidebarPath: require.resolve('./sidebars.js'),
           routeBasePath: '/docs/',
+          showLastUpdateAuthor: true,
+          showLastUpdateTime: true,
           editUrl: ({ docPath }) => {
             // TODO: remove when `latest/` is no longer hardcoded
             const fixedPath = docPath.replace('latest/', '');


Biki-das:
@dsanders11 review

erickzhao (changes requested):
Ran the deploy preview and this doesn't seem to be working in prod.

Not sure if this would work either because the docs files in this repo don't have accurate git history.

Biki-das:
> Ran the deploy preview and this doesn't seem to be working in prod.
> 
> Not sure if this would work either because the docs files in this repo don't have accurate git history.



> 

yeah saw that, is it a infra issue or we can come up with something and make it work?

molant:
@erickzhao does it work if we remove the author?
The date is really useful (and maybe we can get it from the fs), but the author not that much as people come and go and the changes are a bit random at times.

erickzhao:
Seems like this is a caching issue or something. `yarn clear && yarn build && yarn serve` got it working for prod build locally for me.

On the other hand, we ran into the expected problem of this info showing the last update history of the files in this repo, which are inaccurate.

Changing this behaviour would require some wider-scoped refactoring in https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-docs

Biki-das:
> Seems like this is a caching issue or something. `yarn clear && yarn build && yarn serve` got it working for prod build locally for me.
> 
> On the other hand, we ran into the expected problem of this info showing the last update history of the files in this repo, which are inaccurate.
> 
> Changing this behaviour would require some wider-scoped refactoring in https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-docs

so this is the main issue that we are planning to solve in the gsoc project as stated, we are not getting the changes previews so we are not sure of our changes in doc

Biki-das:
> Seems like this is a caching issue or something. `yarn clear && yarn build && yarn serve` got it working for prod build locally for me.
> 
> On the other hand, we ran into the expected problem of this info showing the last update history of the files in this repo, which are inaccurate.
> 
> Changing this behaviour would require some wider-scoped refactoring in https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-docs

yeah local build is fine ,i also gave a proof for the same.

erickzhao:
> so this is the main issue that we are planning to solve in the gsoc project as stated

I don't really understand how you're tying the two issues together. Mind elaborating? 😅

Biki-das:
> > so this is the main issue that we are planning to solve in the gsoc project as stated
> 
> I don't really understand how you're tying the two issues together. Mind elaborating? 😅

ah, sorry to make it complex, so as you we saw in this PR, the deployed preview failed to catch the changes, and we need to manually start the deploy preview as far as i understand, so the project that came into gsoc works around this area itself right?

erickzhao:
Deploy previews I think are automatic for this repo unless they're coming from forks: https://github.com/electron/electronjs.org-new/pull/187

Probably a stretch goal to get that configured, but not our main priority I think. :)

PR #186, opened 2022-03-09
fix: admonitions for crowdin

molant:
This fixes the current spacing with the admonitions so Crowdin doesn't change the format and we can actually build the translated documents.

It also adds a new fixer just in case we don't add the spaces again upstream.


Biki-das (commented):
this change would help the doc with prettier, because prettier tends to make it one liner , if there is no gap between the content and the starting of admonitions and ending of admonitions.

check this out for more references
https://docusaurus.io/docs/markdown-features/admonitions#usage-with-prettier

erickzhao (approved):

PR #187, opened 2022-03-09
feat: new homepage

erickzhao:
Revamps the homepage in various ways. See deploy preview for changes.


Toinane:
Looks really greats! Just one comment about the first CTA button, in dark mode its okay but in light mode, we don't see any border/background and it can be confusing to know if its just a text or a button :
image

Toinane (approved):

src/pages/home/_components/AppsGrid.jsx
```diff
@@ -0,0 +1,43 @@
...
+          >
+            <div className="avatar card__body">
+              <img
+                className={clsx(
+                  'avatar__photo',
+                  styles.appLogo,
+                  isDarkTheme && item.isMonochrome && styles.darkModeInvert
+                )}
+                src={item.image}
+                alt={item.name}
```
Kilian, line 26: It's best to keep this alt here but empty (`alt=""`). Because the image is part of a link, the alt text of the image becomes part of the link, leading to duplication (the link is then read out as eg "Polypane Polypane".) So in this case, keeping it empty helps understanding.

molant:
Looking good!
Couple comments knowing that my goal is to make it easier/faster for people to get started:

- Should the "Try it out" section be closer to the top? Right now it is in the bottom half
- The CTA in the hero is "docs" which might detract people that prefer to jump right away. What do you think of adding a second one "Try it" that points to the "Try it out section" or even www.electronjs.org/fiddle?

BlackHole1 (approved):

BlackHole1:
@erickzhao I noticed that the official website has not been updated. Do we have an expected update date?

erickzhao:
@BlackHole1 we would need to change the Fastly redirects in our legacy Heroku website. Haven't done it yet. :)

BlackHole1:
Do we have any time plan to do this?

erickzhao:
I was originally planning on waiting until we had the full site ready for us to migrate the homepage since it's the most visible one. We have the `/apps` and `/releases` pages left to do, I think.

PR #190, opened 2022-03-12
fix: Fiddle code fix and documentation changes

### LINKED ISSUE #189: Fiddle code for IPC Pattern 3 is buggy and doesn't even work
visit https://www.electronjs.org/docs/latest/tutorial/ipc#pattern-3-main-to-renderer here

Click on Open in Fiddle and try to run the code loaded in fiddle.

**Expected Results**
The code provided should work without errors and warnings.

**Actual Results**
image

image

**Did you want to send a PR ?** Yes

akgupta0777:
Closes #189  

**What I did**
- imported `ipcMain` from electron.
- changed `event.reply` to `event.sender.send` as reply is not a function but sender.send() does the exact same thing.

I verified the fiddle is running and working fine without errors and warnings.

### DIFF
--- a/docs/latest/fiddles/ipc/pattern-3/main.js
+++ b/docs/latest/fiddles/ipc/pattern-3/main.js
@@ -1,4 +1,4 @@
-const {app, BrowserWindow, Menu} = require('electron')
+const {app, BrowserWindow, Menu, ipcMain} = require('electron')
 const path = require('path')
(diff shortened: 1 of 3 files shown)


erickzhao:
This needs to be fixed upstream in the docs: https://github.com/electron/electron/issues/32867

PR #193, opened 2022-03-15
build(deps): update docusaurus to v2.0.0-beta17

erickzhao:


erickzhao:
Looks like the upgrade is blocked on some swizzling I did previously. Will need to make amendments before the upgrade can be merged.

PR #207, opened 2022-04-09
chore: fixing a typo

tawanda-profuse:


tawanda-profuse:
Actually, I'm not sure if it's a chore or a fix but please review...

PR #208, opened 2022-04-09
docs: update taobao to npmmirror for ELECTRON_MIRROR

upupming:
the npm.taobao.org domain will discontinued in 2022.06.30
see the official announcement: https://zhuanlan.zhihu.com/p/465424728


erickzhao:
Should be fixed

PR #211, opened 2022-04-10
fix: add separate openjs logos for themes

jianrong7:
Add two separate logos for light and dark themes for visibility


erickzhao (approved):
LGTM!

PR #240, opened 2022-08-01
Fix broken link to examples in tutorial 4

ariez-xyz:
The link currently displays as follows (see [here](https://www.electronjs.org/docs/latest/tutorial/tutorial-adding-features#how-to-examples)):

    To get started, check out the [How-To Examples][how-to] doc.


molant:
Thanks for fixing this @ariez-xyz !

PR #263, opened 2022-09-22
chore: add some new features and api changes

georgexu99:
title

### DIFF
--- a/blog/electron-21-0.md
+++ b/blog/electron-21-0.md
@@ -25,8 +25,11 @@ The Electron team is excited to announce the release of Electron 21.0.0! You can
 
 ### New Features
 
-TODO: Add features here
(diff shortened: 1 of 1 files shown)


blog/electron-21-0.md
```diff
@@ -26,6 +26,7 @@ The Electron team is excited to announce the release of Electron 21.0.0! You can
 ### New Features
 
 TODO: Add features here
+* Added LoadBrowserProcessSpecificV8Snapshot as a new fuse that will let the main/browser process load its v8 snapshot from a file at browser_v8_context_snapshot.bin. Any other process will use the same path as is used today. [35266](https://github.com/electron/electron/pull/35266)
```
MarshallOfSound, line 29: I don't think we've called out fuses, or any specific fuse in the past.  I don't think this fuse should be used by 99.9% of people so I don't think we should call it out

VerteDinde (approved):
Thanks! 🙇‍♀️

PR #276, opened 2022-10-12
fix(showcase): force `en` locale compare

erickzhao:
Fixed https://github.com/electron/electronjs.org-new/issues/261


BlackHole1 (approved):

BlackHole1 (approved):
It's normal in my local now! Thanks @erickzhao !

PR #296, opened 2022-11-10
fix(api-structures): support reference-style links

### LINKED ISSUE #295: API structure definitions do not work with reference-style links in Markdown
See https://www.electronjs.org/docs/latest/api/ipc-main#ipcmainoncechannel-listener for a live example.

The current AST plugin only uses a find-and-replace algorithm for `link`-type nodes. The "replace" part currently doesn't support definitions.

I'll have to do more research on the AST structure to see if we can do this in an efficient way.

erickzhao:
Fixes #295 

Adds a second pass to the AST walker to find and replace reference-style links. These were previously skipped in the visitor check since we only checked for `link`-type nodes.

For more information on Node types for the Markdown AST, see the [mdast spec](https://github.com/syntax-tree/mdast).

```md

[Example link][1]

[1]: https://example.com

```

### DIFF
--- a/src/components/APIStructurePreview.module.scss
+++ b/src/components/APIStructurePreview.module.scss
@@ -15,6 +15,7 @@
     border: 1px solid var(--tooltip-color);
     border-radius: 20px;
     width: 1.6em;
+    max-height: 55%; // this value should be (1 / line-height) of .markdown p (see custom.scss)
     font-size: 50%;
     text-align: center;
     position: absolute;
--- a/src/transformers/api-structure-previews.js
+++ b/src/transformers/api-structure-previews.js
@@ -2,35 +2,61 @@ const visitParents = require('unist-util-visit-parents');
 const fs = require('fs');
 const path = require('path');
 
-/**
- *
- *
- */
-
 module.exports = function attacher() {
   return transformer;
 };
+
+let structureDefinitions;
+let hasStructures;
+
 /**
  *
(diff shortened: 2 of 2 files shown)


src/transformers/api-structure-previews.js
```diff
@@ -2,45 +2,75 @@ const visitParents = require('unist-util-visit-parents');
...
-  tree.children.unshift({
-    type: 'import',
-    value:
-      "import APIStructurePreview from '@site/src/components/APIStructurePreview';",
-  });
+  structureDefinitions = new Map();
+  hasStructures = false;
+  visitParents(tree, checkLinksandDefinitions, replaceLinkWithPreview);
+  visitParents(tree, isStructureLinkReference, replaceLinkWithPreview);
+  if (hasStructures) {
```
erickzhao, line 21: Orthogonal fix: only import the `APIStructurePreview` component if needed.

src/components/APIStructurePreview.module.scss
```diff
@@ -15,6 +15,7 @@
     border: 1px solid var(--tooltip-color);
     border-radius: 20px;
     width: 1.6em;
+    max-height: 55%; // this value should be (1 / line-height) of .markdown p (see custom.scss)
```
erickzhao, line 18: Orthogonal fix: make this prettier

VerteDinde (approved):

PR #315, opened 2022-11-29
Fixed Delimiter Injection Vulnerability in exportVariable

pavkcon:
## Describe the summary:
The Electron Website provides a set of packages to make creating actions easier. The `core.exportVariable` function uses a well known delimiter that attackers can use to break out of that specific variable and assign values to other arbitrary variables. Workflows that write untrusted values to the `GITHUB_ENV` file may cause the path or other environment variables to be modified without the intention of the workflow or action author. Users should upgrade to `@actions/core v1.9.1`. If you are unable to upgrade the `@actions/core` package, you can modify your action to ensure that any user input does not contain the delimiter `_GitHubActionsFileCommandDelimeter_` before calling `core.exportVariable`.

CVE-2022-35954
[CWE-74](https://cwe.mitre.org/data/definitions/74.html)
[CWE-77](https://cwe.mitre.org/data/definitions/77.html)
`CVSS:3.1/AV:N/AC:L/PR:L/UI:N/S:C/C:N/I:L/A:N`

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,12 +2,13 @@
 # yarn lockfile v1
 
 
-"@actions/core@^1.2.7":
-  version "1.6.0"
-  resolved "https://registry.yarnpkg.com/@actions/core/-/core-1.6.0.tgz#0568e47039bfb6a9170393a73f3b7eb3b22462cb"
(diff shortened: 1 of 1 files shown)


MarshallOfSound:
For security reasons the Electron team does not accept dependency bumps or lockfiles changes from external collaborators.  

I have performed the minimal required bump myself here --> https://github.com/electron/website/pull/317

PR #318, opened 2022-11-29
Improve docs building speed by compressing in parallel ✈️ 

sanjaiyan-dev:
Improving the performance by running the independent `async` functions in parallel ✈️.

_Sorry if I made any mistakes :(_


erickzhao (changes requested):
We currently don't use the versioned docs script and the i18n-build one is going away in #320.

dsanders11:
This PR has conflicts now, and the scripts touched here are being changed by other PRs. Thanks for the PR though!

PR #330, opened 2022-12-01
refactor: convert website to TypeScript

erickzhao:
This pull request rewrites our website in TypeScript, which added first-class support in Docusaurus 2 sometime after we first bootstrapped this project.

The biggest change is that `docusaurus.config.js` becomes a thin shell for `docusaurusConfig.ts`, since Docusaurus doesn't support first-class TypeScript configuration yet (https://github.com/facebook/docusaurus/issues/7911#issuecomment-1207407881).


dsanders11 (approved):
Nice. TIL about the TS `satisfies` keyword. A couple of nit inline comments.

src/transformers/api-options-class.ts
```diff
@@ -0,0 +1,48 @@
...
+    node.data = hastProperties;
+  }
+}
+
+function isParagraph(node: Node): node is Paragraph {
+  return node.type === 'paragraph';
+}
+
+function isOptions(node: Node): node is InlineCode {
+  return node.type === 'inlineCode' && (node as any).value === 'options';
```
dsanders11, line 47:
```suggestion
  return node.type === 'inlineCode' && (node as InlineCode).value === 'options';
```

erickzhao:
gonna need to set up linting for typescript lol cuz we don't have js files in `src/` anymore

erickzhao:
from there, looks like the tests need updating as well.

src/transformers/api-options-class.ts
dsanders11: @erickzhao, why leave the typing as `any` rather than as `InlineCode`? Doesn't the check (`node.type === 'inlineCode') confirm it's of type `InlineCode`?

dsanders11:
If this is still in-progress, can you make it a draft until it's ready for final review? That way it's more clear when that point is reached.

PR #337, opened 2022-12-16
Page update docs/latest/tutorial/tutorial-preload

brendan8c:
I noticed that on the website page (https://www.electronjs.org/docs/latest/tutorial/tutorial-preload)  skipped and didn't specify when to add `const path = require('path')`.  Because of this, an error occurs and the program does not start.


brendan8c:
I await approval and approval for the merger.

erickzhao:
Changes to the raw documentation content should be made in the `electron/electron` docs repo. This website's contents are just a mirror from there.

brendan8c:
@erickzhao, Ok thanks, I'll close that and make it there.

PR #362, opened 2023-01-30
fix(blog): fix npm getting started command

DHainzl:
NPM needs an additional `--` according to the official [Getting Started Docs](https://www.electronforge.io/)

When trying out the "Getting started" command from the blog post I noticed that the webpack plugin was not correctly added. Using the command from the official website works as expected.


erickzhao (approved):
Thank you!

PR #363, opened 2023-01-31
fix: remove old support page from sidebar

dsanders11:
At the moment this is down at the bottom of the sidebar in a duplicated "Tutorial" section, and the page title in the sidebar is awkward, so just remove it all together.

Screen Shot 2023-01-31 at 2 04 12 AM


BlackHole1 (approved):

erickzhao (approved):

erickzhao (changes requested):
Actually, I think this needs to be added to a list so it doesn't get added again.

erickzhao (approved):

PR #364, opened 2023-02-02
Enhancement of container element by aligning to center

Deepu178:
This was the previous alignment:
previous alignment

This is the updated alignment:
added alignment

Please suggest any changes if required.
Thanks

### DIFF
--- a/src/pages/home/_index.module.scss
+++ b/src/pages/home/_index.module.scss
@@ -98,11 +98,3 @@ p {
   padding: 1.2rem;
 }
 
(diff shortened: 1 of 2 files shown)


erickzhao (changes requested):
The left alignment was a conscious design choice. I don't think I particularly like center aligning this.

Deepu178:
Thanks for the feedback
I am looking for  making contribution to the Electron project if you can guide me. That would be nice.

PR #365, opened 2023-02-03
ci: update usage of deprecated ::set-output

dsanders11:
Changed as per these instructions: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


erickzhao (approved):
Thank you!

Issue #369, opened 2023-02-07
Navbar carets should never appear on a separate line


JrDevHarsh:
The items in the navbar menu are not properly aligned center when transitioning from the `mobile` to the `desktop` screen. Here is the screenshot of the bug:

![issue](https://user-images.githubusercontent.com/94737463/217163265-4368f90b-ecf3-4df6-86b9-c67c35c03e0b.png)

JunDevHarsh:
Can I work on this issue?

erickzhao:
Sure thing

santosh-kuchetti:
I would love to contribute to this repo. Could you please assign this issue to me..?

santosh-kuchetti:
Are we having any channel so that we can discuss our doughts..?

JunDevHarsh:
@santosh-kuchetti You can work on this issue if you want to

santosh-kuchetti:
Sure, thanks. Can i know our community channels so that i can disacuss my doughts there

santosh-kuchetti:
the class names that are available in the developer tools are not even available in the code base. It is really hard to find them in the code base.  can you please help me with this @erickzhao?

prachiMeshram:
@erickzhao, I have fixed this issue in my system and created a PR.
Can you please review it and accept the PR

prachiMeshram:
@erickzhao can you please review my PR now

hansupadhyay007:
Hey there..
I really wanted to work on this issue. Could you please assign me this...

MoazMirza-13:
fixed this issue in https://github.com/electron/website/pull/460

MoazMirza-13:
@erickzhao i need someone to review my pr to be merged.

MoazMirza-13:
hey @erikian can you tell me what should i do now ? should i wait for someone to review it ?
and in my notifications i am also seeing that an error occured "check semantic commit", what should i do about it ?

Sankalp2114:
hey i have fixed this issue can you please review it @erickzhao (#530)

PR #373, opened 2023-02-13
navbar carets fixed

prachiMeshram:
Solved issue #369. 
Now navbar carets don't appear on the separate line when transitioning from mobile to desktop.


dsanders11 (changes requested):
This should not modify `package-lock.json` or `yarn.lock`.

PR #375, opened 2023-02-14
fix: Navbar carets should never appear on a separate line

prachiMeshram:
Solved issue #369.
Now navbar carets don't appear on the separate line when transitioning from mobile to desktop.


erickzhao (changes requested):
image
This seems to have caused a regression where menu links are visible when the hamburger menu is present as well.

prachiMeshram:
Sorry for the late response. Was completely occupied with travelling since
last few days.

I will look into this and get back to you.

PR #379, opened 2023-02-24
refactor(gov): generate page on buildtime

erickzhao:
in preparation for a cron build


src/plugins/governance/fetch-data.ts
```diff
@@ -0,0 +1,113 @@
...
+import toString from 'mdast-util-to-string';
+import { Node, Parent } from 'unist';
+import { Table, Link } from 'mdast';
+import * as dotenv from 'dotenv';
+dotenv.config();
+
+export interface WorkingGroup {
+  name: string;
+  description: string;
+  chair: string;
```
dsanders11, line 13: Should we be forward looking and make `chair` optional?

erickzhao: yeah, that makes sense.

erickzhao: 9600d2e

georgexu99 (approved):

dsanders11:
@erickzhao, there's a bug in the merged version of this, see the deployment - it renders with `chair &&` visible on the page.

PR #387, opened 2023-03-15
fix(blog): CEF is Chromium, not Chrome

busybox11:


BlackHole1 (approved):

erickzhao (approved):
oops. thanks for the catch!

PR #389, opened 2023-03-17
chore: remove `<Layout>` props

erickzhao:
These show up as undefined props in our `.tsx` files nowadays.

I think these were removed upstream in https://github.com/facebook/docusaurus/pull/6925

### DIFF
--- a/src/pages/community/index.tsx
+++ b/src/pages/community/index.tsx
@@ -4,7 +4,7 @@ import Link from '@docusaurus/Link';
 
 function Community() {
   return (
-    <Layout title="Community">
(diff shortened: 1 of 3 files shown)


dsanders11:
How do those pages get their HTML `<title>` elements if not from those props?

It also looks like those props are still documented in the Docusaurus docs: https://docusaurus.io/docs/creating-pages#add-a-react-page

erickzhao:
Weirdly, it doesn't. I guess these are needed after all. Let me investigate further about why I'm getting intellisense issues, then.

Issue #408, opened 2023-04-19
API Structures Sidebar Not Alphabetized


dsanders11:
New additions to "API Structures" on the sidebar aren't alphabetized, which is confusing since the large majority of them are, and the other sections like modules are alphabetized.

<img width="285" alt="Screenshot 2023-04-18 at 10 12 53 PM" src="https://user-images.githubusercontent.com/5820654/232972956-4797fef2-5b2f-4166-a8e8-5753a1acd371.png">

abhishekgite446:
Addressed this issue in [PR](https://github.com/electron/website/pull/422)

ChavvaNarayana:
Sounds good to me. I'm willing to contribute.

ashishsoniii:
Hey @dsanders11  , I can see two solutions for this issue:

1. We can sort the items when we are inserting an element into the array, which means changes in [scripts/tasks/create-sidebar.ts](https://github.com/electron/website/blob/main/scripts/tasks/create-sidebar.ts), runs. (It seems like we have to sort the array every time an element is pushed.)

2. Alternatively, we can edit theme/DocPage/Layout/index.tsx and sort the array as we fetch all items from sidebar.js. This way, before passing the array to DocPageLayoutSidebar, we can ensure it's sorted. This seems like a better approach as it avoids changing the script of create-sidebar.js, and the sort function won't run unnecessarily multiple times.

Current data is like this 
```
[ 
Main Process Modules :{Array of links(item) - currently sorting these links}, 
Renderer Process Modules : { },
Custom DOM Elements : { }
Chromium and Node.js : { } , 
Classes: { }
] 

```
So we are already sorting Array of links present inside each element. How about sorting array of Elements also as it is not currently sorted? 

I have implemented and tested all of these solutions, and they are working as expected! Let me know your opinion, and then I will proceed to generate a PR.

botPuneet:
Hey, @dsanders11 i maded a pr based on your guidance pls review the above [pr](https://github.com/electron/website/pull/526)

PR #422, opened 2023-06-27
fix: Alphabetized API Structures sidebar

abhishekgite446:
Fixing the [Issue 408](https://github.com/electron/website/issues/408)

Alphabetized API Structures sidebar ✅


ChavvaNarayana (commented):
It is not considered the correct approach.

abhishekgite446:
Retrieved the hardcoded changes made in the first commit and added a function to sort the sidebar sub-elements in [API tab](https://www.electronjs.org/docs/latest/api/app).


https://github.com/electron/website/assets/253690/fd383962-c553-44a5-95c0-49e18e7df946

dsanders11 (changes requested):
`sidebars.js` is generated by [`scripts/tasks/create-sidebar.ts`](https://github.com/electron/website/blob/main/scripts/tasks/create-sidebar.ts), so it is not the correct place to put these changes.

abhishekgite446:
Hi @dsanders11, thanks for letting me know. I was having a look at the file [scripts/tasks/create-sidebar.ts](https://github.com/electron/website/blob/main/scripts/tasks/create-sidebar.ts) and observed that when I disabled/commented out the **if condition** mentioned below in the same file and executed the command `yarn pre-build `then the [sidebar.js](https://github.com/electron/website/blob/main/sidebars.js) was overwritten with 'API Structures' sorted in alphabetical array. 

I reverted my changes from [scripts/tasks/create-sidebar.ts](https://github.com/electron/website/blob/main/scripts/tasks/create-sidebar.ts) as the sidebar.js is already sorted and going ahead if a new api is added in the folder [docs/latest/api/structures](https://github.com/electron/website/tree/main/docs/latest/api/structures) then the` yarn pre-build` command should add that in [sidebar.js](https://github.com/electron/website/blob/main/sidebars.js) in sorted order.

Would love to know your thoughts on it.
image

PR #432, opened 2023-08-03
style: make `summary` components bold

erickzhao:
ref electron/electron#39337

### DIFF
--- a/src/pages/home/_index.module.scss
+++ b/src/pages/home/_index.module.scss
@@ -3,6 +3,10 @@ p {
   font-size: 1.1rem;
 }
 
+summary {
+  font-weight: bold;
+}
+
 .section {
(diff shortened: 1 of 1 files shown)


dsanders11 (approved):
It's a bit broad, not sure if it might have unintended styling effects. Although did a quick check and we don't use `<summary>` anywhere in the docs, so for the moment it would only affect that. We could potentially scope it to just the target page, since Docusaurus adds a class to `<html>` of the form `docs-doc-id-<URI>`, but that would be fragile if the path changes.

No perfect options here, just thinking out loud.

erickzhao:
Docusaurus does a lot of default styling to `<details>` to make it look in-line with the admonition styling. I'm not too worried about use-cases outside of that... for now.

Issue #433, opened 2023-08-05
[Feature]: Adding hover effect to the FeatureImageContainer in the landing page


RajeevDash007:
Before:
There is :before for every feature image container which is tilted to 5deg which looks good but something should happen when an USER hover on the image to make its UI more cool.

After:
I want to add a :hover:before to 0 deg while hovering with a delay of 0.5s to make it look smooth and cool

I am raising a PR soon.

RajeevDash007:
@dsanders11 whats your view on this?

PR #434, opened 2023-08-05
fix: Added hover effect to Feature Image container: before

RajeevDash007:
#433 Fixed
before: 
https://www.awesomescreenshot.com/video/19701823?key=24d6ec3a9f44b5d67300e80ef120e278

after:
https://www.awesomescreenshot.com/video/19701924?key=ab8aed5db6977604e541ad548a5c1a31


RajeevDash007:
@dsanders11 Check this out

erickzhao (changes requested):
Hi, thanks for this PR, but I don't think a hover effect is suitable here because that usually communicates some clickable intent. I think it's misleading that the object doesn't really lead you anywhere.

PR #462, opened 2023-10-20
Pagination

tabishnaqvi1311:
enhanced the showcase page by adding pagination#453

https://github.com/electron/website/assets/115006878/eee08a46-3425-469e-9273-2255e0b69d9d


erickzhao (changes requested):
Hey @tabishnaqvi1311, thanks for this PR!

Since the apps directory is already lazy-loaded, I don't think the pagination approach is a step forward in usability for the site.

I think we'll have to reject this PR and the associated issue.

PR #471, opened 2023-11-10
fix: version download link prefix

codebytere:
Refs https://github.com/electron/website/pull/469
Closes https://github.com/electron/fiddle/issues/1503

Looks like macOS and Windows lost their `v` version prefix in the shuffle.


codebytere:
Oh lmao missed https://github.com/electron/website/pull/470

PR #482, opened 2023-12-08
feat: add logos of folks providing free infra

MarshallOfSound:
Added the logos of folks giving us free infra


dsanders11 (dismissed):
Something seems off in the preview deployment, the logos are really big?

erickzhao (approved):

PR #488, opened 2023-12-18
feat: tabbed CJS/ESM code blocks

dsanders11:
Combines back-to-back code blocks with language identifiers of CJS followed by MJS into tabbed code blocks. Code blocks can opt-out by adding `no-combine` to the meta string for the code block. Also optionally strips a preamble comment in the code blocks (making them more user-friendly in rendered Markdown not on the website).

No upstream usage in the docs yet. Will add linting on `e/e` to enforce certain things (like the preamble comments).


src/components/ESMCodeBlock.tsx
BlackHole1, line 1: This file contains the handling logic for both esm and cjs, but the file name only reflects esm, which may confuse other maintainers. Perhaps we can change it to: `JSCodeBlock.tsx`

src/transformers/esm-code-blocks.ts
BlackHole1, line 1: Same as above

BlackHole1 (changes requested):

erickzhao (approved):
Agree with @BlackHole1 but otherwise LGTM. Can we also make a demo/playground to give it a spin?

dsanders11:
Renamed files per suggestion and added a demo page (will remove before merge) at `/docs/latest/tabbed-js-code-block/`, [here's the deployed preview.](https://electronjsorg-new-pr-488.herokuapp.com/docs/latest/tabbed-js-code-block/)

BlackHole1 (approved):
LGTM

erickzhao (approved):

PR #493, opened 2024-01-12
chore: make Node.js version more explicit

erickzhao:
Matches what we're building in CI.


package.json
```diff
@@ -3,6 +3,9 @@
   "version": "0.0.0",
   "private": true,
   "license": "Apache-2.0",
+  "engines": {
+    "node": "16"
```
erikian, line 7: I think we want one of these instead:

```suggestion
    "node": ">= 16.0.0"
```
```suggestion
    "node": ">= 16.20.2"
```

erikian (approved):

package.json
erickzhao: It can't hurt to have folks running 18 or 20 if we specify that we use 16 on CI, but I think it would be good for us to keep it to the same major version in engines if we're gonna be strict about it 🤔

erikian: Doesn't `"node": "16"` imply only `16.0.0` though? I guess the syntax I was looking for was `"node": ">=16.0.0 <17"` ([example](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines))

dsanders11:
I know this is just documenting the current reality, but we should upgrade to Node 18+ since 16 is EOL. 🙂

dsanders11 (approved):

codebytere (approved):

PR #494, opened 2024-01-15
build: use Node.js v20 LTS

codebytere:
Follow-up to https://github.com/electron/website/pull/493.


.github/workflows/build-and-deploy.yml
```diff
@@ -32,9 +32,10 @@ jobs:
         run: echo ${{ inputs.branch }}
       # This should be set up from earlier
       - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # tag: v3.3.0
-      - uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c # tag: v3.6.0
+      - name: Setup Node.js v20
```
dsanders11, line 35: nit: I've been removing the version number from text in workflows when updating them (so that it just says "Setup Node.js", etc) so that it's one less thing which needs to be touched on updates (and one less thing which can become incorrect by mistake).

dsanders11 (approved):

ckerr (approved):

PR #496, opened 2024-01-22
ci: add problem matcher for Docusaurus output

dsanders11:
Adds a problem matcher to turn warnings in the Docusaurus build output into warning annotations on the workflow runs so they're more discoverable, and org-wide automation can pick them up.


erickzhao (approved):
TIL this is a feature. Cool!

PR #515, opened 2024-02-17
fix(blog): remove outdated CSS syntax

erickzhao:
I don't think we support this syntax on the website at all anymore (e.g. https://www.electronjs.org/blog/august-2016-roundup), and MDX 3 will explicitly require us to escape these characters in the copy.


VerteDinde (approved):

dsanders11 (approved):
Link check can be ignored since this is touching legacy blog posts and we don't want to open that can of worms.

PR #519, opened 2024-02-21
docs: correct the node.js version for electron 29

selfisekai:
see https://github.com/electron/electron/blob/v29.0.0/DEPS#L6-L7


hardfalcon:
@dsanders11: This appears to still be an issue. Also, the Electron releases page is still stuck with Electron 29, although it should probably go up to Electron 31 or 32 by now:
https://www.electronjs.org/docs/latest/tutorial/electron-timelines
image

Would it be an option to simply replace that page with a link/redirect to https://github.com/electron/electron/blob/main/docs/tutorial/electron-timelines.md?

dsanders11:
@hardfalcon, we're having a known ongoing issue with the website not automatically updating with the latest docs. We're working on fixing it, and have #554 to manually update if we can't get it resolved in a timely manner. Apologies for the inconvenience.

hardfalcon:
Ok, thanks :)

hardfalcon:
@dsanders11: Thanks! :)

PR #526, opened 2024-03-01
API Structures Sidebar Not Alphabetized

botPuneet:
fixes issue #408 added a sorting function in [create-sidebar.ts](https://github.com/electron/website/blob/main/scripts/tasks/create-sidebar.ts) to sort the sidebar in api

### DIFF
--- a/scripts/tasks/create-sidebar.ts
+++ b/scripts/tasks/create-sidebar.ts
@@ -178,6 +178,22 @@ export const createSidebar = async (root: string, destination: string) => {
       prettierConfig
     );
 
+    for (const id of Object.keys(sidebars)) {
+      for (const category of sidebars[id]) {
+        setRecursive(category);
+        // Sort items alphabetically within each category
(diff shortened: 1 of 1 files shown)


botPuneet:
Hey, @dsanders11 pls review this and guide me for any changes it needed

scripts/tasks/create-sidebar.ts
```diff
@@ -178,6 +178,22 @@ export const createSidebar = async (root: string, destination: string) => {
       prettierConfig
     );
 
+    for (const id of Object.keys(sidebars)) {
```
erickzhao, line 181: issue(blocking): this won't work because the `formattedSidebarString` is parsed before we do any of the in-place sorting.

I also think we would want to only have this available for the `API` `Structures` categories, rather than every single sidebar category recursively.

erickzhao (changes requested):

PR #532, opened 2024-03-05
chore: delete unused versions-info.json

erickzhao:
Don't think we use this anymore. It points to 22-x-y and there's no way that's relevant. :)


dsanders11 (approved):
For paper trail, looks like this was added in #120 and then support was removed in #327.

PR #560, opened 2024-05-07
build: upgrade to TypeScript 5.3

erickzhao:
This PR migrates to TypeScript 5, which is a prerequisite for upgrading to Docusaurus 3.

- This adds a major version bump to [`@tsconfig/docusaurus`](https://www.npmjs.com/package/@tsconfig/docusaurus?activeTab=code), which now uses the `bundler` module resolution strategy from TS 5.
- `@typescript-eslint` was upgraded to support TS 5.
  - This is the smallest major version bump that would allow this to work.
  - Note that because this version of `@typescript-eslint` does not support TS 5.4, we're a bit behind in TS versions.
- Removes Jest dependencies since we weren't using them. Can be added back at any point if we add unit tests to the code.

Future work:
- Upgrade ESLint and `@typescript-eslint`
- Upgrade to TS 5.4.5

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,6 @@
     "@electron/lint-roller": "^1.11.1",
     "@tsconfig/docusaurus": "^2.0.3",
     "@types/fs-extra": "^9.0.13",
-    "@types/jest": "^26.0.23",
     "@types/mdast": "^3.0.10",
     "@types/node": "^18.11.10",
     "@types/node-fetch": "^2.6.2",
@@ -104,7 +103,6 @@
     "gray-matter": "^4.0.3",
     "gunzip-maybe": "^1.4.2",
     "husky": "^8.0.0",
-    "jest": "^26.6.3",
     "json5": "^2.2.2",
     "latest-version": "^5.1.0",
     "lint-staged": "^13.2.0",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -104,7 +103,6 @@
     "gray-matter": "^4.0.3",
     "gunzip-maybe": "^1.4.2",
     "husky": "^8.0.0",
-    "jest": "^26.6.3",
```
dsanders11, line 107: Let's remove remove `jest.config.js` and:

https://github.com/electron/website/blob/11a406efd3b82305e7a3dfd3e62845d0142bb5e2/.eslintrc.js#L8

erickzhao: [1b0f806](https://github.com/electron/website/pull/560/commits/1b0f806f1001aa22bafacf98dd7df4dff7a0698b)

tsconfig.json
```diff
@@ -1,11 +1,16 @@
...
-    "types": ["jest"]
+    "target": "ES2015"
+  },
+  "ts-node": {
+    // these options are overrides used only by ts-node
+    // same as the --compilerOptions flag and the TS_NODE_COMPILER_OPTIONS environment variable
+    "compilerOptions": {
+      "module": "commonjs",
+      "moduleResolution": "Node"
+    }
```
erickzhao, line 14: We use `ts-node` specifically to run `docusaurusConfig.ts` from `docusaurus.config.js`.

dsanders11 (approved):

PR #561, opened 2024-05-08
build: React 18

erickzhao:
Necessary for the Docusaurus v3 upgrade.


erickzhao:
This breaks a lot of the Docusaurus 2 peer deps so this should be performed at the same time as the v3 migration

PR #576, opened 2024-06-17
feat: add tags to blog posts

dsanders11:
Stacked on #575.

Adds a `blog/tags.yml` file to define two tags (`release` and `security`) and backfills tags for relevant blog posts. Makes it easier to find blog posts related to those topics, especially since the sidebar only displays the last 50 posts (although we can control this via config).


dsanders11:
"Check Blog Links" failure is due to touching all the old blog posts which have broken links. Can be ignored.

erickzhao (approved):

PR #582, opened 2024-06-21
build: cleanup deps

dsanders11:
* Moves `dotenv` and `execa` to `devDependencies` since they're only used by scripts
* Adds `semver` to `dependencies` since we make use of it in `src/`
* Removes `@actions/core` and `@actions/github` as they went stale in #533


erickzhao (approved):
sweet, thanks

PR #583, opened 2024-06-21
chore: use pngquant to optimize PNG images

dsanders11:
Uses [`imagemin-pngquant`](https://github.com/imagemin/imagemin-pngquant) to optimize PNG images at build time.

Significantly reduces the size of images used on the home page, shrinking transferred size from 4.4 MB to 1.2 MB.

It's technically lossy optimization, but I don't think anyone can tell the difference between [this](https://www.electronjs.org/assets/marketing/app-store.png) and [this](https://electronjsorg-new-pr-583.herokuapp.com/assets/marketing/app-store.png) unless you've got Superman eyes.

**Before**:
Before

**After**:
After


package.json
```diff
@@ -40,6 +40,9 @@
     "docusaurus-plugin-sass": "^0.2.1",
     "dotenv": "^16.0.3",
     "execa": "^5.0.0",
+    "image-minimizer-webpack-plugin": "^4.0.2",
+    "imagemin": "^9.0.0",
+    "imagemin-pngquant": "^10.0.0",
```
BlackHole1, line 45: It seems it should be moved to `devDependencies`.

BlackHole1 (dismissed):

dsanders11:
The compelling reason is the significant reduction in size for imperceptible visual change.

I briefly tied a lossless optimizer on one of the larger files as a test and it was something like a 15% reduction, compared to 75% with pngquant.

erickzhao (approved):
Great idea. LGTM once @BlackHole1's comment is addressed. :)

dsanders11:
Had an offline discussion with @ckerr regarding his comment, and moving ahead with taking the easy win here and we can follow up with other improvements in the future, such as committing losslessly optimized versions of the files to begin with. 👍

PR #586, opened 2024-06-26
ci: temporary workflow to resolve Electron release versions for PRs

dsanders11:
This is a temporary workflow in support of @piotrpdev's Google Summer of Code work. Long-term this will be removed from this repo and the code will be refactored to live on `electron/release-status`, but for now this is the easiest place to put it such that it can be used immediately. Since it is temporary I've kept it isolated entirely to this one file so it doesn't affect anything else in this repo.

Example snippet of how to fetch the content from the latest workflow run is included below. Requires a GitHub token with `actions: read` perms on `electron/website` repo.

```js
const AdmZip = require('adm-zip');

const AUTH_HEADER = { Authorization: 'Bearer ${{ env.GH_TOKEN }}' };

let resp = await fetch('https://api.github.com/repos/electron/website/actions/artifacts', { headers: AUTH_HEADER });
const latestArtifact = (await resp.json()).artifacts.filter(({ name }) => name === 'resolved-pr-versions').sort((a, b) => a.id > b.id)[0];

resp = await fetch(latestArtifact.archive_download_url, { headers: AUTH_HEADER });
const buffer = Buffer.from(await resp.arrayBuffer());

const zip = new AdmZip(buffer);
const data = JSON.parse(zip.readAsText(zip.getEntries()[0]));
```

### DIFF
--- a/.github/workflows/resolve-pr-release-versions.yml
+++ b/.github/workflows/resolve-pr-release-versions.yml
@@ -1,0 +1,262 @@
+name: Resolve PR Release Versions
+
+on:
+  workflow_dispatch:
+    inputs:
+      max_release_count:
+        description: Max number of unprocessed releases to process (all if left blank)
+        required: false
+        type: number
+  schedule:
+    - cron: '0 */6 * * *'
+
+permissions:
+  actions: read
+
+jobs:
+  resolve_pr_release_versions:
+    name: Resolve PR Release Versions
+    runs-on: ubuntu-latest
+    steps:
+      - name: Restore previous run data
+        uses: dawidd6/action-download-artifact@71072fbb1229e1317f1a8de6b04206afb461bd67 # v3.1.2
+        with:
+          name: resolved-pr-versions
+          if_no_artifact_found: ignore
+          workflow_conclusion: 'completed'
+          search_artifacts: true
+      - run: npm install @electron/fiddle-core
+      - uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
+        with:
+          script: |
+            const fs = require('node:fs/promises');
+            const { ElectronVersions } = require('@electron/fiddle-core');
+            const semver = require('semver');
+
+            // https://github.com/electron/trop/blob/a481299dfd522c7b5c5d10e2355ad9e7f0ce193e/src/utils/branch-util.ts#L90-L94
+            const getBackportPattern = () =>
+              /(?:^|\n)(?:manual |manually )?backport (?:of )?(?:#(\d+)|https:\/\/github.com\/.*\/pull\/(\d+))/gim;
+
+            const BOOTSTRAP_DATA_URL = 'https://gist.githubusercontent.com/dsanders11/eb51a04d04a6a3e0710d88db5250e698/raw/fd960b6dea1152b55427407646044f1ba187e52b/data.json';
+            const MIN_MAJOR = 10;
+            const RELEASE_PAGINATION_START = undefined;
+            const RELEASE_MAX_PAGINATION_COUNT = 100;
+            const NEW_RELEASES_QUERY = `query($endCursor: String, $count: Int!) {
+              rateLimit {
+                limit
+                remaining
+                used
+                resetAt
+              }
+              repository(owner: "electron", name: "electron") {
+                releases: refs(
+                  refPrefix: "refs/tags/",
+                  after: $endCursor,
+                  first: $count,
+                  orderBy: {field: TAG_COMMIT_DATE, direction: ASC}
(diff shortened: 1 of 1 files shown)


.github/workflows/resolve-pr-release-versions.yml
```diff
@@ -0,0 +1,262 @@
...
+                      // move on to the next previous release until we're back to a valid release
+                      core.warning(`${previousRelease} is a missing release - skipping`);
+                      previousRelease = versions[--idx];
+                      continue;
+                    }
+
+                    const { compare: { commits } } = release;
+
+                    for (const commit of commits.nodes) {
+                      if (commit.associatedPullRequests.pageInfo.hasNextPage) {
```
erickzhao, line 186: What number does this correspond to?

erickzhao (approved):

.github/workflows/resolve-pr-release-versions.yml
```diff
@@ -0,0 +1,262 @@
...
+            const { ElectronVersions } = require('@electron/fiddle-core');
+            const semver = require('semver');
+
+            // https://github.com/electron/trop/blob/a481299dfd522c7b5c5d10e2355ad9e7f0ce193e/src/utils/branch-util.ts#L90-L94
+            const getBackportPattern = () =>
+              /(?:^|\n)(?:manual |manually )?backport (?:of )?(?:#(\d+)|https:\/\/github.com\/.*\/pull\/(\d+))/gim;
+
+            const BOOTSTRAP_DATA_URL = 'https://gist.githubusercontent.com/dsanders11/eb51a04d04a6a3e0710d88db5250e698/raw/fd960b6dea1152b55427407646044f1ba187e52b/data.json';
+            const MIN_MAJOR = 10;
+            const RELEASE_PAGINATION_START = undefined;
```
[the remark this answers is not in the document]
dsanders11: Nope, good catch. 😄 Originally I was using it to manually set where pagination began, but switched to letting it start from the beginning and using `semver` to ignore releases until `MIN_MAJOR`. Just forgot to remove this code. 👍

.github/workflows/resolve-pr-release-versions.yml
dsanders11: Line 90, `associatedPullRequests(first: 20)`. This check is just a safeguard against a very unlikely edge case, where the number of `associatedPullRequests` might exceed 20 and we wouldn't be able to find the merged one without paginating. I was too lazy to do nested pagination here since it's not actually a problem so long as we keep that number high enough - 20 was arbitrary but is also limited by the overall complexity of the GraphQL query (e.g. if you make `associatedPullRequests` 100 and `labels` 100, the query gets rejected for too high of a potential complexity).

From the docs explaining why there can be more than one `associatedPullRequests`, unfortunately. We only want the merged one, but in theory there can be additional ones associated for release branches so have to handle that case.

> The merged Pull Request that introduced the commit to the repository. If the commit is not present in the default branch, additionally returns open Pull Requests associated with the commit.

.github/workflows/resolve-pr-release-versions.yml
```diff
@@ -0,0 +1,261 @@
...
+          script: |
+            const fs = require('node:fs/promises');
+            const { ElectronVersions } = require('@electron/fiddle-core');
+            const semver = require('semver');
+
+            // https://github.com/electron/trop/blob/a481299dfd522c7b5c5d10e2355ad9e7f0ce193e/src/utils/branch-util.ts#L90-L94
+            const getBackportPattern = () =>
+              /(?:^|\n)(?:manual |manually )?backport (?:of )?(?:#(\d+)|https:\/\/github.com\/.*\/pull\/(\d+))/gim;
+
+            const BOOTSTRAP_DATA_URL = 'https://gist.githubusercontent.com/dsanders11/eb51a04d04a6a3e0710d88db5250e698/raw/fd960b6dea1152b55427407646044f1ba187e52b/data.json';
```
VerteDinde, line 40: I am incredibly nitting here, but do we want to pop this into an Electron repo somewhere? If we think this will actually be pretty temporary, I'm okay leaving it

VerteDinde (approved):
Looks good! One nit question

.github/workflows/resolve-pr-release-versions.yml
dsanders11: It's a valid point to raise! I'm definitely intending for this workflow overall to be pretty temporary (pending practical realities getting int he way), but this line in particular can be super temporary. Once we merge and have one run of the workflow to store an artifact, this line can be removed entirely as the bootstrap step will be complete and shouldn't (unless artifacts somehow all got jettisoned) be needed again.

PR #594, opened 2024-07-02
feat: API History

piotrpdev:
> [!NOTE]
> This PR is part of the **2024 Electron GSoC** [[Project](https://summerofcode.withgoogle.com/programs/2024/projects/wenKR2i3)] [[Proposal](https://gist.github.com/piotrpdev/9ed9bd7f0f3ab192a5ae58a35fbe03e7)].

Use `export GH_TOKEN=<token>` before `yarn start` if you want to test the releases fetching and unzipping.

You might have to manually modify `sidebars.js` to make `yarn start` work.

> [!NOTE]
> In development mode, clicking on one of the breaking changes in the table will open the correct page but not jump to the linked header. This is an issue with Docusaurus.
> - https://github.com/facebook/docusaurus/issues/2399

## TODO

- [x] Remove the changes to `docs/api/*.md` before merge, they're just for previewing/testing the PR.
	- https://github.com/electron/website/pull/594#discussion_r1666038994.
- [x] Remove the test data in `src/transformers/api-history.ts` https://github.com/electron/website/pull/594/files#diff-1150b9be07a8f0d779823f74f0403681feb5bc13bfda0a500edeadd81a8ce61cR94-R113
- ~~Replace fetching the PR release versions artifact with fetching a JSON file once @dsanders11 implements that.~~ Won't happen for a while, will merge with `GH_TOKEN`.
	- https://github.com/electron/website/pull/586


src/transformers/api-history.ts
```diff
@@ -0,0 +1,189 @@
...
+      },
+      42086: {
+        release: 'v32.0.0-nightly.20240531',
+        backports: ['v31.0.0'],
+      },
+    };
+
+    _allPrReleaseVersions = versions;
+    return _allPrReleaseVersions;
+  }
```
piotrpdev, line 64: @dsanders11 Should I make the `LOCAL_DEV` case default for now so it shows up in the Heroku deploys?

dsanders11: @piotrpdev, for the immediate future that might be the best short-term solution. I don't know enough about the Heroku preview deployments to know if we can provide it with a GH token, I'll look into it.

src/css/custom.scss
```diff
@@ -256,3 +256,25 @@
...
+    &:hover {
+      cursor: pointer;
+    }
+  }
+
+  table {
+    margin-top: 0.5rem;
+    margin-bottom: 0;
+  }
+}
```
dsanders11, line 280: As discussed on our call, let's move these to `src/components/ApiHistoryTable.module.scss` to match the pattern we're using with `src/components/APIStructurePreview.tsx`.

src/transformers/api-history.ts
```diff
@@ -0,0 +1,189 @@
...
+  function maybeGenerateApiHistoryTable(
+    node: Code,
+    ancestors: Parent[]
+  ): ActionTuple | void {
+    const parent = ancestors[0];
+    const idx = parent.children.indexOf(node);
+
+    const apiHistory = parseYaml(node.value) as ApiHistory;
+
+    const versionsInHistory = [];
```
dsanders11, line 139:
```suggestion
    const prsInHistory = [];
```

Current name is a bit inaccurate.

src/components/ApiHistoryTable.tsx
```diff
@@ -0,0 +1,111 @@
...
+function generateTableRow(
+  prReleaseVersions: PrReleaseVersionsContainer,
+  type: Change,
+  prUrl: string,
+  changes?: string
+) {
+  const prNumber = prUrl.split('/').at(-1);
+
+  const releaseStatus = prReleaseVersions[Number(prNumber)];
+  // TODO: Sorting
```
dsanders11, line 31: You can make use of `compareVersions` from `@electron/fiddle-core` here.

src/transformers/api-history.ts
piotrpdev: Made default in [`fd76258` (#594)](https://github.com/electron/website/pull/594/commits/fd7625898ccd26bd48da1022b7208946a54b2d05) for now.

src/css/custom.scss
piotrpdev: Resolved in [`d9e15cb` (#594)](https://github.com/electron/website/pull/594/commits/d9e15cb575ea94a5061566127e9bec1c2d3958ca)

src/transformers/api-history.ts
piotrpdev: Resolved in [`c927a72` (#594)](https://github.com/electron/website/pull/594/commits/c927a722b773b90c355da10f0d0519b0697469c2)

src/components/ApiHistoryTable.tsx
piotrpdev: The table will only include released major versions so I'm not sure if that's necessary. Can just do:

```js
[ "30.0.0", "29.0.0", "31.0.0" ].toSorted((a, b) => b.split(".")[0] - a.split(".")[0])
```

src/transformers/api-history.ts
```diff
@@ -0,0 +1,262 @@
...
+  const archiveDownloadResponse = await fetch(
+    latestArtifact.archive_download_url,
+    fetchOptions
+  );
+  const buffer = Buffer.from(await archiveDownloadResponse.arrayBuffer());
+
+  const zip = new AdmZip(buffer);
+  const parsedData = JSON.parse(
+    zip.readAsText(zip.getEntries()[0]!)
+  ) as PrReleaseArtifact;
```
piotrpdev, line 116: > @erickzhao https://github.com/electron/lint-roller/pull/73#discussion_r1670619581
> nit: Would it be possible to expand this line into a few more to make it easier to debug if something goes wrong?
> 
> For example:
> 
> * Checking `Array.isArray` and `length` instead of non-null coercing
> * An `isPRReleaseArtifact` type guard function instead of a type assertion

piotrpdev: Resolved in [`394cc14` (#594)](https://github.com/electron/website/pull/594/commits/394cc144af0fd1c1938304be8ea982cfb5f56ede)

scripts/tasks/preprocess-api-history.ts
```diff
@@ -0,0 +1,198 @@
...
+const getMarkdownFiles = async (startPath: string) => {
+  const filesPaths = await globby(path.posix.join(startPath, '**/*.md'));
+
+  const files: Map<string, string> = new Map();
+  for (const filePath of filesPaths) {
+    const content = await readFile(filePath, 'utf-8');
+    files.set(filePath, content);
+  }
+
+  return files;
```
BlackHole1, line 41: Use concurrency to achieve better performance.
```suggestion
  const filePromises: Promise<[string, string]>[] = filesPaths.map(f => {
    return fs.readFile(f, "utf-8").then(content => [f, content]);
  });
  
  const contents = await Promise.all(filePromises);
  return new Map(contents);
```

scripts/tasks/preprocess-api-history.ts
```diff
@@ -0,0 +1,198 @@
...
+
+// Copied from here but slightly different: <https://github.com/electron/lint-roller/blob/bac245478ba69017c2a82e8ba1b2884a37647494/bin/lint-markdown-api-history.ts#L42-L67>
+function findPossibleApiHistoryBlocks(tree: Root) {
+  const codeBlocks: Html[] = [];
+
+  visit(
+    tree,
+    (node: Node) =>
+      node.type === 'html' &&
+      (node as LiteralString).value.toLowerCase().includes('```') &&
```
BlackHole1, line 52:
```suggestion
      (node as LiteralString).value.includes('```') &&
```

src/transformers/api-history.ts
```diff
@@ -0,0 +1,314 @@
...
+    ('added' in value || 'deprecated' in value || 'changes' in value)
+  );
+}
+
+export default function attacher() {
+  return transformer;
+}
+
+function matchApiHistoryCodeBlock(node: Node): node is Code {
+  return isCode(node) && node.lang === 'YAML' && node.meta === 'history';
```
BlackHole1, line 63: I found that this PR presents multiple inconsistent behaviors when handling `node`. Is this expected? As follows:
1. `node.lang === 'YAML'`
2. `toLowerCase().includes('YAML') === 'YAML'` and `lang?.toLowerCase()`
3. `node.meta === 'history'`
4. `toLowerCase().includes('history') === 'history'`

Some places use `toLowerCase`, while others do not. Can you standardize the processing method?

BlackHole1 (changes requested):

scripts/tasks/preprocess-api-history.ts
piotrpdev: I measured the performance for that suggestion and got a significantly slower result:

```
238 Markdown documents
~30ms		current
100-300ms	suggestion
```

We also decided to reduce the number of Markdown files we check in [`247b7d9` (#594)](https://github.com/electron/website/pull/594/commits/247b7d964e309538371c0a26631061832580ed23). Same trend in that scenario:

```
70 Markdown documents
~10ms		current
40-100ms	suggestion
```

scripts/tasks/preprocess-api-history.ts
```diff
@@ -0,0 +1,231 @@
...
+
+function findValidApiHistoryBlocks(
+  possibleHistoryBlocks: Html[],
+  filePath: string,
+  validateAgainstSchema: ValidateFunction<ApiHistory>
+) {
+  const validHistoryBlocks: Html[] = [];
+
+  // All of the validation logic copied from here: <https://github.com/electron/lint-roller/blob/bac245478ba69017c2a82e8ba1b2884a37647494/bin/lint-markdown-api-history.ts#L117-L176>
+  for (const possibleHistoryBlock of possibleHistoryBlocks) {
```
erickzhao, line 71: For all of the `continue` clauses in this loop, can we add some line/column numbers based on `possibleHistoryBlock.position`?

I imagine it would be hard to find the exact failure spot for an API file as big as `app.md` for example.

scripts/tasks/preprocess-api-history.ts
```diff
@@ -0,0 +1,199 @@
...
+
+  visit(
+    tree,
+    (node: Node) =>
+      node.type === 'html' &&
+      (node as LiteralString).value.includes('```') &&
+      (node as LiteralString).value.toLowerCase().includes('yaml') &&
+      (node as LiteralString).value.toLowerCase().includes('history'),
+    (node) => {
+      codeBlocks.push(node as Html);
```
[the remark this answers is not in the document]
piotrpdev: Is that really necessary in this case? That part of the code only runs if `node.type === 'html'` is true so we **know** it's of type `Html` _and_ that's the only use of `node` in that scope.

scripts/tasks/preprocess-api-history.ts
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    validHistoryBlocks.push(possibleHistoryBlock);
+  }
+
+  return validHistoryBlocks;
+}
+
+export const preprocessApiHistory = async (startPath: string) => {
+  // ? Allow for a custom schema to be passed in.
```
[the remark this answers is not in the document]
piotrpdev: It's a question i.e. maybe we should allow for a custom schema? Felt like that was up for discussion,

package.json
```diff
@@ -37,16 +37,24 @@
...
     "@docusaurus/remark-plugin-npm2yarn": "3.4.0",
     "@docusaurus/theme-mermaid": "3.4.0",
+    "@electron/fiddle-core": "^1.3.3",
     "@mdx-js/react": "^3.0.0",
+    "adm-zip": "^0.5.14",
+    "ajv": "^8.16.0",
     "clsx": "^1.1.1",
     "docusaurus-plugin-sass": "^0.2.1",
+    "dotenv": "^16.0.3",
+    "execa": "^5.0.0",
```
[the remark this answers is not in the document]
piotrpdev: Whoops, good catch! Pretty sure that's leftover from a merge conflict. Fixed in [`99b36d1` (#594)](https://github.com/electron/website/pull/594/commits/99b36d14fb6dc17615bf82923cc223cee0ae7ce9).

package.json
```diff
@@ -37,16 +37,24 @@
...
+    "hast-util-from-html": "^2.0.1",
+    "mdast-util-from-markdown": "^2.0.1",
```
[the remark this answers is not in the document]
piotrpdev: Good point, moved to `devDeps` in [`51ab304` (#594)](https://github.com/electron/website/pull/594/commits/51ab3045c05d62aefa77266cfd223bbc8e4c1561). Both of these are only used in the pre-process script.

scripts/tasks/preprocess-api-history.ts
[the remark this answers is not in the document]
piotrpdev: Added in [`9481b1c` (#594)](https://github.com/electron/website/pull/594/commits/9481b1caa53605d2e37e6b478c66f0807f17b565)

scripts/tasks/preprocess-api-history.ts
piotrpdev: Isn't this script meant to just check the block is valid and process it? If you want more details aren't you meant to use the `lint-roller` script?

scripts/tasks/preprocess-api-history.ts
BlackHole1: Oh, that's interesting, using concurrency actually made it slower.😅🤔

BlackHole1: I just tried it locally, and the result was the opposite of yours. Can you double-check? Here is my test code:

```js
import globby from 'globby';
import path from 'path';
import { readFile } from 'fs/promises';

const startPath = 'docs/latest';

const filesPaths = await globby(path.posix.join(startPath, 'api', '/*.md'));

console.log('count:', filesPaths.length);

const getMarkdownFiles = async () => {
  const files = new Map();
  for (const filePath of filesPaths) {
    const content = await readFile(filePath, 'utf-8');
    files.set(filePath, content);
  }

  return files;
};

const getMarkdownFiles2 = async () => {
  const ps = filesPaths.map((f) => {
    return readFile(f, 'utf-8').then((content) => [f, content]);
  });

  const contents = await Promise.all(ps);
  return new Map(contents);
};

console.time('getMarkdownFiles1');
await getMarkdownFiles();
console.timeEnd('getMarkdownFiles1');

console.time('getMarkdownFiles2');
await getMarkdownFiles2();
console.timeEnd('getMarkdownFiles2');
```

Result:

```
$ node ./test.mjs
count: 72
getMarkdownFiles1: 12.383ms
getMarkdownFiles2: 5.1ms
```

My machine info: `MacBook Pro Intel Core i9 2.4GHz 32GB`

scripts/tasks/preprocess-api-history.ts
```diff
@@ -0,0 +1,232 @@
...
+  visit(
+    tree,
+    (node: Node) =>
+      isHtml(node) &&
+      (node as LiteralString).value.includes('```') &&
+      (node as LiteralString).value.toLowerCase().includes('yaml') &&
+      (node as LiteralString).value.toLowerCase().includes('history'),
+    (node) => {
+      if (!isHtml(node)) return; // Should never happen, but TypeScript needs this check
+      codeBlocks.push(node);
```
BlackHole1, line 58:
```suggestion
      codeBlocks.push(node as Html);
```

scripts/tasks/preprocess-api-history.ts
```diff
@@ -0,0 +1,232 @@
...
+          filePath
+        )}`
+      );
+      continue;
+    }
+
+    if (
+      codeBlock.type !== 'code' ||
+      codeBlock.lang?.toLowerCase() !== 'yaml' ||
+      codeBlock.meta?.trim() !== 'history'
```
BlackHole1, line 112: Do need add `toLowerCase`?

scripts/tasks/preprocess-api-history.ts
```diff
@@ -0,0 +1,232 @@
...
+}
+
+export const preprocessApiHistory = async (startPath: string) => {
+  const schema = path.resolve(startPath, 'api-history.schema.json');
+
+  let validateAgainstSchema: ValidateFunction<ApiHistory> | null = null;
+
+  try {
+    const ajv = new Ajv();
+    const ApiHistorySchemaFile = await readFile(schema, { encoding: 'utf-8' });
```
BlackHole1, line 159: Maintain a consistent code style.

```suggestion
    const ApiHistorySchemaFile = await readFile(schema, 'utf-8');
```

scripts/tasks/preprocess-api-history.ts
```diff
@@ -0,0 +1,232 @@
...
+    validateAgainstSchema = ajv.compile(ApiHistorySchema);
+  } catch (error) {
+    logger.error(`Error reading API history schema:\n${error}`);
+    return;
+  }
+
+  const files = await getMarkdownFiles(startPath);
+
+  for (const [filePath, content] of files) {
+    if (!content.includes('<!--')) continue;
```
BlackHole1, line 172: Can we move it inside the `getMarkdownFiles` function? The data producer should ensure that their data is accurate and should not require the consumer to perform data cleaning again.

BlackHole1 (changes requested):

scripts/tasks/preprocess-api-history.ts
```diff
@@ -0,0 +1,232 @@
...
+      );
+
+      // Stripping the HTML comment tags of a history block will offset the position of the next history block
+      fileLengthDifference +=
+        historyBlockWithoutTags!.length - validHistoryBlock.value.length;
+
+      newContent = start + historyBlockWithoutTags + end;
+    }
+
+    await writeFile(filePath, newContent, { encoding: 'utf-8' });
```
BlackHole1, line 230:
```suggestion
    await writeFile(filePath, newContent, 'utf-8');
```

src/components/ApiHistoryTable.tsx
```diff
@@ -0,0 +1,163 @@
...
+  prReleaseVersionsJson: string;
+}
+
+function generateTableRow(
+  prReleaseVersions: PrReleaseVersions | undefined,
+  type: Change,
+  change:
+    | NonNullable<ApiHistory['added']>[0]
+    | NonNullable<ApiHistory['deprecated']>[0]
+    | NonNullable<ApiHistory['changes']>[0]
```
BlackHole1, line 29:
```suggestion
  change: ApiHistory[keyof ApiHistory][0]
```

src/components/ApiHistoryTable.tsx
```diff
@@ -0,0 +1,163 @@
...
+) {
+  const prNumber = Number(change['pr-url'].split('/').at(-1));
```
BlackHole1, line 31: Maybe we should add a `prNumber` parameter to this function?

BlackHole1 (changes requested):
I found that in the current changes, we only need the number of the PR, not the whole URL of the PR.
Moreover, the code has a lot of handling of URLs (fetching the PR number through URLs), so can we consider changing `pr-url` to `pr-id`?
> I might be missing some context, so this issue might have been discussed before, but I'm not sure. Please feel free to provide additional information if needed.

PTAL @erickzhao @dsanders11 @VerteDinde

scripts/tasks/preprocess-api-history.ts
piotrpdev: Whoops, fixed in [`a111aea` (#594)](https://github.com/electron/website/pull/594/commits/a111aeaccb88bcfc0abd44dd923ae8d546292cf7)

scripts/tasks/preprocess-api-history.ts
piotrpdev: Hmm, I tried your test code and got the same result as you. I'm probably just using the profiler wrong then:

```patch
diff --git a/scripts/tasks/preprocess-api-history.ts b/scripts/tasks/preprocess-api-history.ts
index fb95451..5cbc3af 100644
--- a/scripts/tasks/preprocess-api-history.ts
+++ b/scripts/tasks/preprocess-api-history.ts
@@ -33,13 +33,12 @@ interface LiteralString extends Literal {
 const getMarkdownFiles = async (startPath: string) => {
   const filesPaths = await globby(path.posix.join(startPath, 'api', '/*.md'));
 
-  const files: Map<string, string> = new Map();
-  for (const filePath of filesPaths) {
-    const content = await readFile(filePath, 'utf-8');
-    files.set(filePath, content);
-  }
+  const filePromises: Promise<[string, string]>[] = filesPaths.map(f => {
+    return readFile(f, 'utf-8').then((content) => [f, content]);
+  });
 
-  return files;
+  const contents = await Promise.all(filePromises);
+  return new Map(contents);
 };
 
 // Copied from here but slightly different: <https://github.com/electron/lint-roller/blob/bac245478ba69017c2a82e8ba1b2884a37647494/bin/lint-markdown-api-history.ts#L42-L67>
@@ -166,7 +165,9 @@ export const preprocessApiHistory = async (startPath: string) => {
     return;
   }
 
+  console.profile('preprocessApiHistory');
   const files = await getMarkdownFiles(startPath);
+  console.profileEnd('preprocessApiHistory');
 
   for (const [filePath, content] of files) {
     if (!content.includes('<!--')) continue;

```

| Original | Suggestion |
|:---------:|:------------:|
| Screenshot 2024-07-16 154133 | Screenshot 2024-07-16 154416 |

src/components/ApiHistoryTable.tsx
piotrpdev: I had it that way before ¯\\_(ツ)_/¯

```patch
    ...(apiHistory.changes?.map((change) => {
      const prNumber = Number(change['pr-url'].split('/').at(-1));
      return generateTableRow(
+		prNumber 
        prReleaseVersions[prNumber],
        Change.CHANGED,
        change
      );
    }) ?? []),
```

I personally prefer the way it is currently.

scripts/tasks/preprocess-api-history.ts
piotrpdev: If my profiling is wrong and you're right though, it's only like a `~5ms` difference. Maybe at that point it's better to use the current code since it's easier to debug? @dsanders11

scripts/tasks/preprocess-api-history.ts
piotrpdev: Hmm, are you sure? I think `getMarkdownFiles` should just get Markdown files. I don't think it should be responsible for filtering by file contents too. @dsanders11 @erickzhao @VerteDinde ?

scripts/tasks/preprocess-api-history.ts
piotrpdev: Decided to revert in Slack meeting, [`6e74b23` (#594)](https://github.com/electron/website/pull/594/commits/6e74b234bd6c901a78cc54870e56639db9dc0ed9)

scripts/tasks/preprocess-api-history.ts
piotrpdev: Discussed in Slack meeting, resolved in [`04029d1` (#594)](https://github.com/electron/website/pull/594/commits/04029d12875b8862a2139c1839a107f7df1c318d)

erickzhao: As discussed in the sync today, we should do two things to make the error message clearer:

* Clarify that the `Error` will simply skip the block rather than fail parsing.
* Tell users to run `lint-roller` to resolve any warnings.

piotrpdev: > As discussed in the sync today, we should do two things to make the error message clearer:
> 
> * Clarify that the `Error` will simply skip the block rather than fail parsing.
> * Tell users to run `lint-roller` to resolve any warnings.

Sorry, was this a delayed reply? I think [`04029d1` (#594)](https://github.com/electron/website/pull/594/commits/04029d12875b8862a2139c1839a107f7df1c318d) implements what you mentioned here.

scripts/tasks/preprocess-api-history.ts
dsanders11: Let's stick with the current code. It was a good suggestion from @BlackHole1 to try out, but given the mixed results and minor potential difference, I don't think it's worth pursuing here.

BlackHole1: OK, I can accept both:)

scripts/tasks/preprocess-api-history.ts
BlackHole1: When I conduct a code review, I would assume that the purpose of `getMarkdownFiles` is: "to return the correct list of md files." Therefore, in my view, the function name should be something like: `findMarkdownsWithAPIHistory`. The main reason for this feeling is that this function is not used elsewhere; it serves only for `api-history`.

If we go by the function name `getMarkdownFiles`, your approach is fine :)

src/components/ApiHistoryTable.tsx
BlackHole1: Because I found a lot of code like `Number(added['pr-url'].split('/').at(-1))` appearing, I raised this question to see if there's a better way to optimize this code.

Of course, over-encapsulation and optimization are sometimes not a good idea. So, this is just a personal opinion :)

> BTW, regarding code style, I'm just giving my opinion. The final decision still lies with the person doing this.

BlackHole1:
Before answering this question, I need to apologize first. Due to work and time zone differences, I wasn't able to participate in the RFC discussions and Slack discussions, which caused me to miss a lot of context.

----

> @BlackHole1 Sorry, can you be a bit more specific on where you want to change pr-url to pr-id? Are you talking about making that change in the schema (https://github.com/electron/rfcs/pull/6), or just in some of the code in this PR?

I recently discovered during the code review of this PR that the current changes almost do not use the pr-url, but only use the pr-id within it. To get the pr id, there are many lines of code like `Number(added['pr-url'].split('/').at(-1))`. I believe this is redundant code.
I looked through https://github.com/electron/rfcs/pull/6, but did not see any relevant discussion, so I am raising this question here.

> Using URLs also doesn't restrict us to just electron/electron which could be useful in the future.

Although I don't see when support will be extended to repositories outside of electron/electron in the short term, I can reluctantly accept this reason.

> It's also useful in scenarios like these:
https://github.com/electron/website/pull/594/files#diff-1dd7a747a324f6f8916a65d87d4c435b9041ed1b0a0d9fa48518423b4d3012edR42-R57

Yes, pr-url is only used here. My question is, why not use pr-id directly and then use something like: `https://github.com/electron/electron/pulls/${prID}` here? Wouldn't this reduce the redundancy in the code overall?

----

Overall, this PR, CODE LGTM. Since other members think pr-url is acceptable, I won't strongly oppose it, just raising a question :)

BlackHole1 (approved):
My question has been stated, but it should not be used as a reason to block the merging of the PR, as the code looks good.
If others have additional comments on my question, they are also welcome to leave comments :)

src/components/ApiHistoryTable.tsx
```diff
@@ -0,0 +1,160 @@
...
+        prReleaseVersions[prNumber],
+        Change.CHANGED,
+        change
+      );
+    }) ?? []),
+    ...(apiHistory.added?.map((added) => {
+      const prNumber = Number(added['pr-url'].split('/').at(-1));
+      return generateTableRow(prReleaseVersions[prNumber], Change.ADDED, added);
+    }) ?? []),
+  ];
```
dsanders11, line 140:
```suggestion
  const generateChangeRows = (changeType, changes) => {
    return (changes ?? []).map((change) => {
      const prNumber = Number(change['pr-url'].split('/').at(-1));
      return generateTableRow(
        prReleaseVersions[prNumber],
        changeType,
        change
      );
    });
  };

  const apiHistoryChangeRows = [
    ...generateChangeRows(Change.DEPRECATED, apiHistory.deprecated),
    ...generateChangeRows(Change.CHANGED, apiHistory.changes),
    ...generateChangeRows(Change.ADDED, apiHistory.added),
  ];
```

IMO this reads a lot cleaner if you pull out the common code. Type annotations left as an exercise for the reader. 🙂

src/components/ApiHistoryTable.tsx
```diff
@@ -0,0 +1,160 @@
...
+        {type}
+      </pre>
+    );
+  }
+
+  if ('breaking-changes-header' in change) {
+    changesJsx = (
+      <a
+        href={
+          // ? Is this okay? Is there an existing function that should be used here to generate the URL?
```
dsanders11, line 84: I think this is fine.

src/components/ApiHistoryTable.tsx
```diff
@@ -0,0 +1,160 @@
+import React from 'react';
+import { Details } from '@docusaurus/theme-common/Details';
+import {
+  ApiHistory,
+  PrReleaseVersions,
+  PrReleaseVersionsContainer,
+} from '../transformers/api-history';
+import ReactMarkdown from 'react-markdown';
+
+import styles from './ApiHistoryTable.module.scss';
```
dsanders11, line 10:
```suggestion
import { Details } from '@docusaurus/theme-common/Details';
import React from 'react';
import ReactMarkdown from 'react-markdown';

import {
  ApiHistory,
  PrReleaseVersions,
  PrReleaseVersionsContainer,
} from '../transformers/api-history';
import styles from './ApiHistoryTable.module.scss';
```

Sorting imports is fun.

src/transformers/api-history.ts
```diff
@@ -0,0 +1,318 @@
...
+  const zipEntries = zip.getEntries();
+  const firstZipEntry = zipEntries[0];
+
+  if (firstZipEntry == null)
+    throw new Error('No entries found in the artifact archive.');
+
+  const zipText = zip.readAsText(firstZipEntry);
+  const parsedData = JSON.parse(zipText);
+  if (!isPrReleaseArtifact(parsedData))
+    throw new Error('Invalid PR release artifact.');
```
dsanders11, line 166:
```suggestion
  if (!isPrReleaseArtifact(parsedData)) {
    throw new Error('Invalid PR release artifact.');
  }
```

Let's avoid this style, it's error-prone.

src/transformers/api-history.ts
```diff
@@ -0,0 +1,318 @@
...
+  const archiveDownloadResponseBuffer =
+    await archiveDownloadResponse.arrayBuffer();
+  const archiveDownloadBuffer = Buffer.from(archiveDownloadResponseBuffer);
+
+  const zip = new AdmZip(archiveDownloadBuffer);
+  const zipEntries = zip.getEntries();
+  const firstZipEntry = zipEntries[0];
+
+  if (firstZipEntry == null)
+    throw new Error('No entries found in the artifact archive.');
```
dsanders11, line 161:
```suggestion
  if (firstZipEntry == null) {
    throw new Error('No entries found in the artifact archive.');
  }
```

Let's avoid this style, it's error-prone.

src/transformers/api-history.ts
```diff
@@ -0,0 +1,318 @@
...
+
+  if (!isGithubArtifactsListResponse(artifactsListResponseJson))
+    throw new Error('Invalid GitHub artifacts list response.');
+
+  const latestArtifact = artifactsListResponseJson.artifacts
+    .filter(({ name }) => name === 'resolved-pr-versions')
+    .sort((a, b) => b.id - a.id)[0];
+
+  if (!latestArtifact)
+    throw new Error('No resolved-pr-versions artifact found.');
```
dsanders11, line 146:
```suggestion
  if (!latestArtifact) {
    throw new Error('No resolved-pr-versions artifact found.');
  }
```

Let's avoid this style, it's error-prone.

src/transformers/api-history.ts
```diff
@@ -0,0 +1,318 @@
...
+  };
+
+  const artifactsListResponse = await fetch(
+    GH_ACTIONS_ARTIFACTS_URL,
+    fetchOptions
+  );
+  const artifactsListResponseJson = await artifactsListResponse.json();
+
+  if (!isGithubArtifactsListResponse(artifactsListResponseJson))
+    throw new Error('Invalid GitHub artifacts list response.');
```
dsanders11, line 139:
```suggestion
  if (!isGithubArtifactsListResponse(artifactsListResponseJson)) {
    throw new Error('Invalid GitHub artifacts list response.');
  }
```

Let's avoid this style, it's error-prone.

src/transformers/api-history.ts
```diff
@@ -0,0 +1,318 @@
+import logger from '@docusaurus/logger';
+import { Code, Node } from 'mdast';
+import { getJSXImport, isCode, isImport } from '../util/mdx-utils';
+import { ActionTuple, SKIP, visitParents } from 'unist-util-visit-parents';
+import { ElectronVersions, SemVer } from '@electron/fiddle-core';
+import { Parent } from 'unist';
+import { parse as parseYaml } from 'yaml';
+import semver from 'semver';
+import AdmZip from 'adm-zip';
```
dsanders11, line 9:
```suggestion
import logger from '@docusaurus/logger';
import { ElectronVersions, SemVer } from '@electron/fiddle-core';
import AdmZip from 'adm-zip';
import { Code, Node } from 'mdast';
import semver from 'semver';
import { Parent } from 'unist';
import { ActionTuple, SKIP, visitParents } from 'unist-util-visit-parents';
import { parse as parseYaml } from 'yaml';

import { getJSXImport, isCode, isImport } from '../util/mdx-utils';
```

scripts/tasks/preprocess-api-history.ts
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    await writeFile(filePath, newContent, 'utf-8');
+  }
+
+  if (hasWarned) {
+    logger.warn(
+      'Some API history blocks were skipped due to errors. ' +
+        'Run "lint-roller-markdown-api-history" from "electron/lint-roller" for more details.'
+    );
```
dsanders11, line 245:
```suggestion
    logger.warn('Some API history blocks were skipped due to errors.');
```

I don't think we need to mention the linter here, given it's removed from the process and is ingesting docs which should have already passed linting.

src/transformers/api-history.ts
```diff
@@ -0,0 +1,318 @@
...
+async function getAllElectronVersions(): Promise<SemVer[]> {
+  if (_allElectronVersions) {
+    return _allElectronVersions;
+  }
+
+  const { versions } = await ElectronVersions.create(undefined, {
+    ignoreCache: true,
+  });
+
+  if (versions == null) throw new Error('No Electron versions found.');
```
dsanders11, line 81: I don't think this check does anything, `versions` is always an array. Not sure we really need to check for this, an error should be thrown if something goes wrong with `ElectronVersions.create`.

src/transformers/api-history.ts
```diff
@@ -0,0 +1,318 @@
...
+
+  _allElectronVersions = versions;
+  return _allElectronVersions;
+}
+
+let _allPrReleaseVersions: PrReleaseVersionsContainer;
```
dsanders11, line 87:
```suggestion
let _allPrReleaseVersions: PrReleaseVersionsContainer | undefined;
```

More accurate typing.

src/transformers/api-history.ts
```diff
@@ -0,0 +1,318 @@
...
+
+function matchApiHistoryCodeBlock(node: Node): node is Code {
+  return (
+    isCode(node) &&
+    node.lang?.toLowerCase() === 'yaml' &&
+    node.meta?.toLowerCase() === 'history'
+  );
+}
+
+let _allElectronVersions: SemVer[];
```
dsanders11, line 70:
```suggestion
let _allElectronVersions: SemVer[] | undefined;
```

More accurate typing.

scripts/tasks/preprocess-api-history.ts
```diff
@@ -0,0 +1,247 @@
...
+import path from 'path';
+import { fromMarkdown } from 'mdast-util-from-markdown';
+import { visit } from 'unist-util-visit';
+import type { Html, Root } from 'mdast';
+import type { Node, Literal } from 'unist';
+import Ajv, { JSONSchemaType, ValidateFunction } from 'ajv';
+import { readFile, writeFile } from 'fs/promises';
+import { fromHtml } from 'hast-util-from-html';
+import logger from '@docusaurus/logger';
+import { parse as parseYaml } from 'yaml';
```
dsanders11, line 11:
```suggestion
import logger from '@docusaurus/logger';
import Ajv, { JSONSchemaType, ValidateFunction } from 'ajv';
import { readFile, writeFile } from 'fs/promises';
import { fromHtml } from 'hast-util-from-html';
import globby from 'globby';
import type { Html, Root } from 'mdast';
import { fromMarkdown } from 'mdast-util-from-markdown';
import path from 'path';
import type { Node, Literal } from 'unist';
import { visit } from 'unist-util-visit';
import { parse as parseYaml } from 'yaml';
```

dsanders11 (changes requested):

scripts/tasks/preprocess-api-history.ts
piotrpdev: @erickzhao Asked for this in the July 16th meeting, but alright.

src/components/ApiHistoryTable.tsx
piotrpdev:
```suggestion
```

src/components/ApiHistoryTable.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+  const apiHistory = JSON.parse(apiHistoryJson) as ApiHistory;
+  const prReleaseVersions = JSON.parse(
+    prReleaseVersionsJson
+  ) as PrReleaseVersionsContainer;
+
+  // ? Maybe this is too much abstraction?
+  // ? `added` and `deprecated` go on the ends, but is there any point in sorting `changes`?
+  //  How do you sort them? Given change X and change Y, X could have been merged into main before Y,
+  //  but Y could have been backported to a stable version before X.
```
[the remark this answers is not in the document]
piotrpdev:
```suggestion
```

src/components/ApiHistoryTable.tsx
piotrpdev: Looks great, good idea 👍

src/transformers/api-history.ts
piotrpdev: Sure, would you mind explaining/link to some resources as to how/why it is error prone?

src/transformers/api-history.ts
piotrpdev:
```suggestion
```

src/transformers/api-history.ts
piotrpdev: Also, I'm assuming you didn't write this code then? 😆

https://github.com/electron/website/blob/76c112bea9a8c00d887696fced27bda0d3c76021/.github/workflows/resolve-pr-release-versions.yml#L250

piotrpdev: Should I change these too?

https://github.com/electron/website/pull/594/files#diff-1dd7a747a324f6f8916a65d87d4c435b9041ed1b0a0d9fa48518423b4d3012edR34-R37

```ts
  const release = prReleaseVersions?.release;
  if (release) allVersions.push(release);

  const backports = prReleaseVersions?.backports;
  if (backports) allVersions.push(...backports);
```

erickzhao (approved):

src/components/ApiHistoryTable.tsx
```diff
@@ -49,12 +55,20 @@ function generateTableRow(
...
           {version}
         </pre>
       </a>
     );
-  });
+
+    formattedVersions.push(formattedVersion);
+
+    // If backport is a major (i.e. >=x.0.0), no need to include the release or any other backports
+    if (isBackportMajor) break;
```
dsanders11, line 67: Is skipping "any other backports" correct behavior here? I think we still want to display the case where it was backported to an older branch, while also consolidating the backport majors.

E.g., if it landed on `main`, then was backported to `31-x-y` while that branch was still in beta, and also backported to `30-x-y`, the release versions for the PR might look like 32.0.0 (main release), 31.0.0 (backport release), and 30.x.y. In this case we'd want the API history to display: ">=31.0.0, ^30.x.y". I think with breaking here it would only display ">=31.0.0"?

piotrpdev: Man, @MarshallOfSound, I like this feature and thank you for suggesting it. But I feel like I need a degree in theoretical physics just to implement/debug/discuss it correctly 😭.

@dsanders11 Due to the witchcraft/black magic necessary to implement this, I had to mess around with the sorting of the releases/backports. Are you sure you didn't miss these changes when you were trying to visualize the hell spawn that is this feature?

[`a047f6a` (#594)](https://github.com/electron/website/pull/594/commits/a047f6a2b6f09ee4d6a3950a95ff516643902a2d#diff-1dd7a747a324f6f8916a65d87d4c435b9041ed1b0a0d9fa48518423b4d3012edR39-R40)

(We now iterate from older to newer, i.e. 29, 30, etc. So, in your example, the `break;` would happen on `30.x.y`.

```patch
-  // Sort by major version number e.g. 30.0.0 -> 30 in descending order i.e. 30, 29, ...
-  allVersions.sort((a, b) => Number(b.split('.')[0]) - Number(a.split('.')[0]));
+  // Sort by major version number e.g. 30.0.0 -> 30 in ascending order i.e. 29, 30, ...
+  allVersions.sort((a, b) => Number(a.split('.')[0]) - Number(b.split('.')[0]));
```

[`a047f6a` (#594)](https://github.com/electron/website/pull/594/commits/a047f6a2b6f09ee4d6a3950a95ff516643902a2d#diff-1dd7a747a324f6f8916a65d87d4c435b9041ed1b0a0d9fa48518423b4d3012edR70-R71)

(We then reverse the order so the table display newest on the top i.e. 30, 29, ...)

```patch
+  // Reverse the order of the versions so that the latest version is first i.e. 30, 29, ...
+  formattedVersions.reverse();
```

dsanders11: I think you're right @piotrpdev, it looks like it's working as intended, I think the comment is just a bit confusing. The way I originally read "no need to include the release" made it sound like it was not including the current release, but I now see it means "no need to include the **main** release". I think the comment would be easier to understand if it said something like "no need to check newer releases".

It looks like `semver` has a function that sounds like it would be helpful here, `simplifyRange`, but I tested it out and it doesn't produce the output we'd want. It seems to favor turning `^` into `x - y` style ranges.

src/components/ApiHistoryTable.tsx
```diff
@@ -0,0 +1,160 @@
...
+  prReleaseVersionsJson: string;
+}
+
+function generateTableRow(
+  prReleaseVersions: PrReleaseVersions | undefined,
+  type: Change,
+  change: NonNullable<ApiHistory[keyof ApiHistory]>[0]
+) {
+  const prNumber = Number(change['pr-url'].split('/').at(-1));
+  const key = prNumber + '-' + type;
```
dsanders11, line 29:
```suggestion
  const key = change['pr-url'] + '-' + type;
```

Unless keys can't have special characters like slashes or colons, seems unnecessary to extract the PR number.

src/components/ApiHistoryTable.tsx
dsanders11: I played around with the code a bit and I think some restructuring can make it less confusing. 🙂

```diff
diff --git a/src/components/ApiHistoryTable.tsx b/src/components/ApiHistoryTable.tsx
index 728f133..9456df1 100644
--- a/src/components/ApiHistoryTable.tsx
+++ b/src/components/ApiHistoryTable.tsx
@@ -1,6 +1,7 @@
 import { Details } from '@docusaurus/theme-common/Details';
 import React from 'react';
 import ReactMarkdown from 'react-markdown';
+import * as semver from 'semver';
 
 import {
   ApiHistory,
@@ -28,47 +29,30 @@ function generateTableRow(
   const prNumber = Number(change['pr-url'].split('/').at(-1));
   const key = prNumber + '-' + type;
 
-  const allVersions: Array<string> = [];
+  // Map versions to semver syntax - use a caret unless it was released in
+  // an X.0.0 release, which gets >= since every release after that has it
+  const versionRanges: Array<string> = semver
+    .rsort(prReleaseVersions?.backports ?? [])
+    .map((version) => (version.endsWith('.0.0') ? '>=' : '^') + version);
 
+  // Only include the main release if it wasn't backported and released in
+  // an X.0.0 release. This consolidates ranges like >=30.0.0 || >=29.0.0
+  // into a single >=29.0.0 since that's more intuitive for developers.
   const release = prReleaseVersions?.release;
-  if (release) allVersions.push(release);
-
-  const backports = prReleaseVersions?.backports;
-  if (backports) allVersions.push(...backports);
-
-  // Sort by major version number e.g. 30.0.0 -> 30 in ascending order i.e. 29, 30, ...
-  allVersions.sort((a, b) => Number(a.split('.')[0]) - Number(b.split('.')[0]));
-
-  const formattedVersions: JSX.Element[] = [];
-
-  for (const version of allVersions) {
-    const [, minor, patch] = version.split('.');
-    const isBackportMajor =
-      release !== version && minor === '0' && patch === '0';
-
-    const formattedVersion = (
-      <a
-        key={version}
-        href={change['pr-url']}
-        target="_blank"
-        rel="noopener noreferrer"
-      >
-        {/* Semver shenanigans, feature backported to both ^7.1.0 and ^6.3.0 would not be present in 7.0.0 */}
-        <pre>
-          {release === version || isBackportMajor ? '>=' : '^'}
-          {version}
-        </pre>
-      </a>
-    );
-
-    formattedVersions.push(formattedVersion);
-
-    // If backport is a major (i.e. >=x.0.0), no need to include the release or any other backports
-    if (isBackportMajor) break;
+  if (release && !versionRanges.find((version) => version.startsWith('>='))) {
+    versionRanges.unshift(`>=${release}`);
   }
 
-  // Reverse the order of the versions so that the latest version is first i.e. 30, 29, ...
-  formattedVersions.reverse();
+  const formattedVersions = versionRanges.map((version) => (
+    <a
+      key={version}
+      href={change['pr-url']}
+      target="_blank"
+      rel="noopener noreferrer"
+    >
+      <pre>{version}</pre>
+    </a>
+  ));
 
   let changesJsx: JSX.Element | undefined;
 
```

dsanders11 (commented):
Added a couple of suggestions to simplify the code.

src/components/ApiHistoryTable.tsx
piotrpdev: Patch applied in [`e61e7d6` (#594)](https://github.com/electron/website/pull/594/commits/e61e7d6fee589a7a4b6dcede8664bc867bee6875)

dsanders11 (approved):

dsanders11:
I think this is ready. 🎉 @piotrpdev, can you go ahead and remove the `docs/` changes? Then I'll go ahead and merge.

piotrpdev:
> @dsanders11
> I think this is ready. 🎉 @piotrpdev, can you go ahead and remove the `docs/` changes? Then I'll go ahead and merge.

Done, hopefully nothing breaks on merge 🤞

PR #608, opened 2024-08-01
chore: use markdownlint-cli2 directly for linting Markdown

dsanders11:
`electron-markdownlint` will be removed in v3 of `@electron/lint-roller`, and using `markdownlint-cli2` allows our custom rules to be used by the `markdownlint` VS Code extension.

This also flips on some of the custom rules we turned on for `e/e` regarding the MDX 3 transition, to ensure going forward that the blog posts don't have any syntax trouble. Required updating some of the blog posts to fix newlines in link text - nothing was broken since the Docusaurus bug that avoids only affects relative links, but the custom rule is generic so it catches all cases.


dsanders11:
Blog link check failure is from touching old blog posts.

erickzhao (approved):
To clarify, we're going to migrate https://github.com/electron/lint-roller/blob/e392deea117053b81f962423e0a0033cfbb9754f/bin/markdownlint-cli-wrapper.ts to `markdownlint-cli2` as well?

dsanders11:
> To clarify, we're going to migrate https://github.com/electron/lint-roller/blob/e392deea117053b81f962423e0a0033cfbb9754f/bin/markdownlint-cli-wrapper.ts to `markdownlint-cli2` as well?

That file will be removed in v3, this is migrating ahead of time. That file was only necessary as an easy way to inject the custom rules instead of forcing users of `electron-markdownlint` to use a CLI option to add them. `markdownlint-cli2` supports the `customRules` field in the config file itself. I can actually probably push that up into the default config now (haven't looked at it yet).

PR #622, opened 2024-08-12
build: docusaurus 3.5 upgrade

erickzhao:
Updates to Docusaurus 3.5.

* Reverts 1f8a694331c4d7be7c86528a1ebb5b6af6648271 since the feature was upstreamed. Removed `patch-package` since we don't need patches anymore.
* Added `mdast-util-mdx` to the dependency tree so that we can pass it in as an extension to `toMarkdown()`. We do some hacky Markdown -> MDAST -> Markdown conversion in the API structure preview transformer that now has an MDX node by default as of https://github.com/facebook/docusaurus/pull/10335 so this is needed to render those previews now.
* Changed an internal import that was updated in the latest version.


alicelovescake (approved):
Thanks for cleaning up my patches!

dsanders11:
This got merged as I was writing this up. 😅

Looks like there's a few visual changes to the blog as a result of this PR. Not to bikeshed too much, but I kind prefer some parts of the blog sidebar as they currently are, like the border below the title. I wonder if we can retain any of those styles without needing to swizzle the whole component?

The blog author name not being centered vertically [looks to be intentional](https://github.com/facebook/docusaurus/blame/140570bdc4adb61f3ed33c6e237aa562c6742370/packages/docusaurus-theme-classic/src/theme/Blog/Components/Author/index.tsx#L90-L93) but IMO is awkward looking. Looks like if we used the new authors pages features and put author's GitHub as a social it might render better.

**Before:**

Before


**After:**

After

PR #623, opened 2024-08-13
style: prettier blog sidebar styles

erickzhao:
Minimal lines of code to achieve the same effect I think?

Ref https://github.com/electron/website/pull/622#issuecomment-2285035680

Upside: no swizzling
Downside: these attribute selectors are brittle and Docusaurus considers them to be internal implementation details: https://docusaurus.io/docs/styling-layout#theme-class-names

### DIFF
--- a/src/css/custom.scss
+++ b/src/css/custom.scss
@@ -264,3 +264,17 @@
 .navbar__logo {
   width: 2rem;
 }
+
+// blog sidebar custom styles
+[class*='BlogSidebar'] {
+  &[class*='yearGroupHeading_'] {
+    margin: 0.75rem 0 0.5rem;
+    color: var(--subtle);
+    border-bottom: 0.01rem solid var(--ifm-table-border-color);
+    padding-bottom: 0.2rem;
+  }
+
+  &[class*='sidebar_'] {
+    padding: 0.5rem;
+  }
+}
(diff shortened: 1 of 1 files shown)


src/css/custom.scss
```diff
@@ -264,3 +264,14 @@
...
+// blog sidebar custom styles
+[aria-label*='Blog recent posts navigation'] {
+  padding: 0.5rem;
+  & [class*='yearGroupHeading_'] {
+    margin: 0.75rem 0 0.5rem;
+    color: var(--subtle);
+    border-bottom: 0.01rem solid var(--ifm-table-border-color);
+    padding-bottom: 0.2rem;
+  }
+}
```
dsanders11, line 277:
```suggestion
.plugin-blog [class*='sidebar_'] {
  padding: 0.5rem;
  & [class*='yearGroupHeading_'] {
    margin: 0.75rem 0 0.5rem;
    color: var(--subtle);
    border-bottom: 0.01rem solid var(--ifm-table-border-color);
    padding-bottom: 0.2rem;
  }
}
```

I don't love using the `aria-label` as a selector, that's going to be pretty fragile. Looks like the blog page has a `plugin-blog` class added to the `html` element, let's use that instead along with the `sidebar_` class.

dsanders11 (changes requested):

dsanders11 (approved):

PR #630, opened 2024-08-20
chore: update docs to 32-x-y

erickzhao:
The `electron-website-updater` hook seems to be failing recently based on the intermittent failures from the service logging.

Here's a manual bump in the meantime so we can get everything running for the upcoming official release.

I just ran `yarn pre-build` locally.


VerteDinde (approved):

sidebars.js
```diff
@@ -189,6 +189,7 @@ module.exports = {
             'latest/development/patches',
             'latest/development/source-code-directory-structure',
             'latest/development/testing',
+            'latest/development/api-history-migration-guide',
```
dsanders11, line 192: nit: these are otherwise sorted alphabetically, do we want to do the same with this one?

erickzhao:
Needs an additional sidebar fix, see Heroku logs

PR #683, opened 2024-11-21
chore: add bsky account link

erickzhao:
https://bsky.app/profile/electronjs.org


dsanders11 (approved):
Do we want to drop the X link right below it at the same time?

PR #690, opened 2024-11-28
refactor: reduce usage of markdown splicing

### LINKED ISSUE #682: Incorrect rendering of embedded API structure hover cards
Image

### LINKED ISSUE #691: Links point to the wrong place if markdown files have the same name
Well well well, if it isn't the consequences of our own assumptions.

https://github.com/electron/website/blob/9487f624415d8a4950e33acb4e2142aad8b75db2/scripts/tasks/md-fixers.ts#L351-L359

We initially implemented the above map to fix a bunch of potentially misplaced links from upstream. However, this causes a problem where we get collisions for documents with the same basename.

This breaks things such as the first link in https://www.electronjs.org/docs/latest/tutorial/navigation-history, which is supposed to point to the API class according to the raw documentation on `electron/electron`.

erickzhao:
Fixes #691
Fixes #682

## Background

This pull request simplifies some of the pre-processing steps that we use to transform the raw documentation in `electron/electron` into the rendered HTML on the website.

```mermaid
graph TD
  F[Download raw documentation from Electron repository] --> A[Regex-Based Markdown Processing]
  A --> B[Custom MDX Transformer Plugins]
  B --> C[docusaurus-plugin-content-docs]

  subgraph E[run as pre-build npm script]
    A
    F
  end

  subgraph D[run within Docusaurus]
    B
    C
  end
```

The first pre-processing step involves making changes to the raw Markdown files on disk by using regex-based replacement in `yarn pre-build`. The replacements made are crude but necessary for edge cases that can't be parsed by Docusaurus' MDX parser.

For example, MDX's usage of JSX means that `` tags need to be self-closing, while they don't need to be according to the HTML spec:

https://github.com/electron/website/blob/9487f624415d8a4950e33acb4e2142aad8b75db2/scripts/tasks/md-fixers.ts#L86-L97

## Problem 

The `fixLinks` function attempts to normalize in the documentation by storing a hashmap of each file and its respective Markdown string contents. It then replaces every single Markdown link with the normalized link matching the file name. This storage mechanism causes problems when two or more documents have the same name under different paths (e.g. API documentation and a tutorial for some exported module).

This fixer function was added early in the website's development cycle (#3) and is useless for the vast majority of links in the docs.

However, `fixLinks` is necessary to get another one of our Markdown hacks working. Namely, `inlineApiStructures` adds the ability to insert API structures in other documents using an `?inline` Markdown link.

This situation leaves us at an impasse where we have broken links in our docs with `fixLinks` enabled, but we can't fix the problem without breaking links in `inlineApiStructures`.

## Solution

This PR tackles the problem by migrating the inlining process to Docusaurus' MDX rendering pipeline, which parses the raw Markdown into an AST and converts it into HTML.

At a high level, we already have multiple plugins that manipulate ASTs, one of which (`api-structure-previews`) stores API structure content in memory to render hover previews on API pages.

This PR reuses this architecture to separately inline the appropriate API structure content in the AST and removes `fixLinks` and `inlineApiStructures` entirely.

## Additional notes

* `api-structure-previews` now stores the MDAST for each structure instead of the raw string to facilitate inlining.
  * This change allows the MDX plugin to avoid converting MDAST -> String -> HTML. For this reason, we don't need the `react-markdown` library anymore.
  * Instead, we directly process MDAST -> HAST -> HTML, which is rendered in React using `dangerouslySetInnerHTML`.
* During PR review, please skip the entire `/docs/` directory. It switches all normalized links to their raw counterparts, which _still work_. Docusaurus does not throw any broken link errors on build times.


VerteDinde (approved):
Thanks for fixing these! 🙇‍♀️

src/components/APIStructurePreview.tsx
```diff
@@ -94,23 +94,23 @@ interface CardProps {
...
+      ref={props.innerRef}
+      id="structures-tooltip"
+      role="tooltip"
+      className={clsx('alert', 'alert--info', 'shadow--md', styles.card)}
+      style={props.position}
+    >
+      <div
+        className="card__body"
+        dangerouslySetInnerHTML={{
+          __html: toHtml(toHast(JSON.parse(props.content))),
```
dsanders11, line 109: [According to the `mdast` docs on security](https://github.com/syntax-tree/mdast?tab=readme-ov-file#security), we should run the result of `toHast` through `hast-util-sanitize` before running it through `toHtml`.

src/components/APIStructurePreview.tsx
```diff
@@ -94,23 +94,23 @@ interface CardProps {
...
-  <article
-    ref={props.innerRef}
-    id="structures-tooltip"
-    role="tooltip"
-    className={clsx('alert', 'alert--info', 'shadow--md', styles.card)}
-    style={props.position}
-  >
-    <div className="card__body">
-      <ReactMarkdown
-        allowedElements={['h1', 'ul', 'li', 'code', 'a', 'em']}
```
dsanders11, line 107: I wonder if we can retain this behavior, to ensure we don't introduce a regression. [Here's the relevant code](https://github.com/remarkjs/react-markdown/blob/9eb589e828445916dfb521117040d8d5420a5e9d/lib/index.js#L258-L278) in `react-markdown`, which simplifies when you exclude the options we wouldn't use (`allowElement` and `unwrapDisallowed`).

dsanders11:
> For example, MDX's usage of JSX means that `` tags need to be self-closing, while they don't need to be according to the HTML spec:

This is a tangent, but that feels like a good candidate that we might be able to fix upstream with a linting rule.

dsanders11 (changes requested):
Partial review, just wanted to throw the requested changes on now since it's a security-related change. Will finish reviewing the rest tomorrow.

src/components/APIStructurePreview.tsx
erickzhao: [a149300](https://github.com/electron/website/pull/690/commits/a1493006b38992102b2337694aa6d7f805175b85)

src/components/APIStructurePreview.tsx
erickzhao: Good point. Here's a regression I found for example:

- https://electronjs.org/docs/latest/api/web-contents/#contentsgetprintersasync
- https://am-i-cooking-chat.electron-website.pages.dev/docs/latest/api/web-contents/#contentsgetprintersasync

erickzhao: [2e44fbb](https://github.com/electron/website/pull/690/commits/2e44fbb4819d0e99d505ae331d02ee8574ebc14b)

src/transformers/api-structure-previews.ts
```diff
@@ -1,48 +1,253 @@
...
+                      return h('a', { href }, [text]);
+                    } else if (node.name === 'header') {
+                      const text = toString(node);
+                      return h('h1', [text]);
+                    }
+                  },
+                },
+              });
+
+              const sanitized = sanitize(HAST, {
```
dsanders11, line 179: Let's add a more detailed comment before this line.

src/transformers/api-structure-previews.ts
```diff
@@ -1,48 +1,253 @@
...
+            if (isInline) {
+              // we inline the structure content as the last sibling of the current node
+              const siblings = parents[parents.length - 1].children;
+              const filtered = filter(
+                structureContent,
+                (node) => node.type !== 'heading'
+              );
+              siblings.push(filtered);
+            } else {
+              const HAST = toHast(structureContent as Root, {
```
dsanders11, line 162: Let's add a more detailed comment before this line.

dsanders11 (commented):
Overall LGTM, I'm not able to find any regressions. Cases checked:
* ✅ Inlines
  * ✅ Labels render
  * ✅ Hovers on referenced structures work
  * ✅ Nested inlines unfurl correctly
* ✅ Hover previews
  * ✅ First header renders correctly
  * ✅ Labels render
  * ✅ Referenced structures are plain links and not previews on previews

Leaving two comments where I think we should add more detailed comments for posterity (you did a great job of commenting other parts of the code and improving the existing comments!), then ready to merge!

src/transformers/api-structure-previews.ts
erickzhao: https://github.com/electron/website/pull/690/commits/c3d0a0eb3522c14c356049aa011af8688cee800a

src/transformers/api-structure-previews.ts
erickzhao: https://github.com/electron/website/pull/690/commits/c3d0a0eb3522c14c356049aa011af8688cee800a

dsanders11 (approved):
Great refactor! :shipit:

PR #703, opened 2024-12-11
build(deps): docusaurus 3.6.3 + faster

erickzhao:
Upgrades to latest Docusaurus and swaps out the underlying build toolchain.

* Installs `@docusaurus/faster`, which is an experimental opt-in change that swaps webpack for rspack amongst other tooling changes.
* Removes `esbuild-loader`, which we were using to run esbuild bundling via webpack.

Anecdotally, cold builds seem to have gone from 60s long to 30s long on my M1 Max MacBook.


dsanders11:
This looks interesting. Can we rebase this on `main` once #709 is merged? That's the only custom Webpack build step we have AFAIK. I know the Docusaurus blog post says most Webpack plugins should work with `rspack`, but might as well do that PR first since it ensures this is cleaner.

erickzhao:
@dsanders11 Done. The other custom part that I removed here was using `esbuild-loader` instead of whatever the default would be for Docusaurus.

dsanders11 (approved):

PR #710, opened 2024-12-12
build: update prettier and eslint-plugin-prettier

dsanders11:
As in title.


dsanders11:
"Check Blog Links" check can be ignored since this touched old blog posts with broken links.

package.json
```diff
@@ -92,7 +92,7 @@
     "mdast-util-mdx": "^3.0.0",
     "mdast-util-to-hast": "^13.2.0",
     "mdast-util-to-string": "^4.0.0",
-    "prettier": "^2.8.0",
+    "prettier": "^3.4.2",
```
erickzhao, line 95: That's a whole lot of commas. Should we preserve existing `trailing-commas` settings to avoid a huge diff or just keep it as-is?

dsanders11: I'd vote keep it as-is, this makes it consistent with other repos.

erickzhao (approved):

PR #738, opened 2025-01-17
docs: Why Electron?

felixrieseberg:
So, why the heck even use Electron? Doesn't the operating system already ship with a web view? And in any case, aren't all good apps native apps?

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -5,6 +5,7 @@ module.exports = {
       label: 'Get Started',
       items: [
         'latest/tutorial/introduction',
+        'latest/why-electron',
         {
           type: 'category',
           label: 'Tutorial',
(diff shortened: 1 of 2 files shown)


sidebars.js
```diff
@@ -3,11 +3,7 @@ module.exports = {
     {
       type: 'category',
       label: 'Get Started',
-      items: [
-        'latest/tutorial/introduction',
-        'latest/tutorial/quick-start',
```
felixrieseberg, line 8: We no longer have the quick start, so this had to go!

erickzhao: @felixrieseberg can we add this doc to the sidebar as well under `Get Started`?

erickzhao (approved):

PR #742, opened 2025-01-18
fix: avoid leaking `h2` styles globally

erickzhao:
Some styles from #676 were leaking because Docusaurus exports a single giant CSS file for consumption (ref https://github.com/facebook/docusaurus/issues/6228).

By querying `h2` directly, this rule was applying to every `h2` tag that didn't have any higher specificity rule attached to it.


dsanders11 (approved):

dsanders11:
Not sure why the Cloudflare preview didn't show up for this PR, but LGTM so I'm gonna YOLO it.

Issue #749, opened 2025-01-27
Inline code baseline issues


nikwen:
The baseline for inline code is weird on the Electron docs website. Words are placed "too low".

![Image](https://github.com/user-attachments/assets/9dccb8c4-0bca-4f6e-b389-499b4a887277)

[Source](https://www.electronjs.org/docs/latest/api/utility-process#event-error-experimental)

Browser: Firefox 134.0.2 on macOS 15.2

dv2202:
i would want to work on this

nikwen:
@dv2202 Please go ahead and open a pull request!

vishalshahh:
Can you assign this task to me

dv2202:
https://github.com/electron/website/pull/771

PR #768, opened 2025-03-02
Fix inline code baseline alignment issue

Ayush9026:
### PR Description

This PR fixes the inline code baseline alignment issue on the Electron documentation website. The issue causes inline code snippets (<code> elements) to appear misaligned (too low) in Firefox 134.0.2 on macOS 15.

Closes https://github.com/electron/website/issues/749

### DIFF
--- a/src/css/custom.scss
+++ b/src/css/custom.scss
@@ -179,10 +179,14 @@
     --ifm-h5-font-size: 0.8rem;
   }
 
-  // We want this to only apply to inline code, so don't apply
-  // this color to ``` code blocks nor any headings.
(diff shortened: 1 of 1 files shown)


erickzhao (changes requested):
I checked out this PR, and it seems like the problem is actually _worse_, at least on Chrome?

| electronjs.org | This PR |
|----------------|---------|
| www electronjs org_docs_latest_api_utility-process | localhost_3000_docs_latest_api_utility-process (1) |

Also, the unrelated changes added along with the added comments (and the one we had that was removed!) makes it seem to me like this PR is copy/pasted from ChatGPT. Please don't do this in the future.

PR #771, opened 2025-03-05
style: fix inline code baseline alignment

### LINKED ISSUE #749: Inline code baseline issues

dv2202:
Summary
This PR fixes the issue where inline code (code elements) had a misaligned baseline on the Electron docs website when viewed on Mozilla 134.0.2 macOS.

Changes
Adjusted CSS styles to ensure proper alignment of inline code elements.
Verified the fix across different browsers to maintain consistency.

Testing
Tested on Mozilla 134.0.2 macOS ✅
Verified on Chrome and Safari to ensure no regressions ✅

Fixes #749

### DIFF
--- a/src/css/custom.scss
+++ b/src/css/custom.scss
@@ -284,3 +284,7 @@
   height: 24px;
   width: 24px;
 }
+
+code {
+  vertical-align: baseline;
+}
--- a/src/css/custom.scss
+++ b/src/css/custom.scss
@@ -179,6 +179,11 @@
     --ifm-h5-font-size: 0.8rem;
   }
 
+  // Fixes inline code alignment in Firefox
+  code {
+    vertical-align: baseline;
+  }
+
   // We want this to only apply to inline code, so don't apply
   // this color to ``` code blocks nor any headings.
   :not(pre):not(h2):not(h3):not(h4):not(h5):not(h6) > code {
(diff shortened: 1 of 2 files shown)


erickzhao (changes requested):
Hmm, I pulled down this change locally and I'm not sure I see a positive change? Can you show a before/after screenshot with the alignment improvements?

I don't think this change should go in the `_index.module.scss` code either, since it's unintuitively leaking a global style from a CSS Module file.

A change like this would belong more in https://github.com/electron/website/blob/main/src/css/custom.scss for example.

nikwen:
`vertical-align: baseline` fixes the issue for me in Firefox on macOS.

However, it needs to be applied in the correct file (see previous comment).

Then it overwrites [Infima's default](https://github.com/facebookincubator/infima/blob/5890e170c2299cab7deaa877f812fa735e60b149/packages/core/styles/content/code.pcss#L31) `vertical-align: middle`.

### Before

image

### After

image

erickzhao:
Ah, it didn't seem to change anything on Chrome. Happy to accept this as as FF-only PR!

dv2202:
I migrated the styles from `index.module.css` to `custom.scss` to improve maintainability and consistency. Tested the changes, and everything is working as expected. Let me know if any further modifications are needed!

PR #807, opened 2025-05-09
fix wdio/playwright code

tkomde:
#### Description of Change

While investigating the E2E test, I found some problems with the documentation, so I created this PR. The following corrections have been made.

(1) wdio sample code may be old and not functional

Both of the original codes (accessing Electron APIs) seemed to not work due to API changes on the WDIO side. I have replaced it with the sample from their documentation (https://webdriver.io/docs/desktop-testing/electron/api/).

The original code can be made to work by doing the following, but it is not a good sample, so I am replacing it.

```
import { browser, expect, $ } from "@wdio/globals";

describe("when the make smaller button is clicked", () => {
  it("should decrease the window height and width by 10 pixels", async () => {
    const boundsBefore = await browser.electron.execute((electron) => {
      const win = electron.BrowserWindow.getFocusedWindow();
      return win.getBounds();
    });

    expect(boundsBefore.width).toEqual(210);
    expect(boundsBefore.height).toEqual(310);

    await $(".make-smaller").click();

    const boundsAfter = await browser.electron.execute((electron) => {
      const win = electron.BrowserWindow.getFocusedWindow();
      return win.getBounds();
    });
    expect(boundsAfter.width).toEqual(200);
    expect(boundsAfter.height).toEqual(300);
  });
});
```

(2) playwright code optimization

- Comfirmed working with latest playwright version (1.52.0).
- Changed require to import.
- The entry points are changed to “.” as in the last sample code.

#### Checklist

### DIFF
--- a/docs/latest/tutorial/automated-testing.md
+++ b/docs/latest/tutorial/automated-testing.md
@@ -83,44 +83,21 @@ Furthermore, WebdriverIO allows you to access Electron APIs to get static inform
 ```js @ts-nocheck
 import { browser, $, expect } from '@wdio/globals'
 
-describe('when the make smaller button is clicked', () => {
-  it('should decrease the window height and width by 10 pixels', async () => {
-    const boundsBefore = await browser.electron.browserWindow('getBounds')
-    expect(boundsBefore.width).toEqual(210)
-    expect(boundsBefore.height).toEqual(310)
-
-    await $('.make-smaller').click()
-    const boundsAfter = await browser.electron.browserWindow('getBounds')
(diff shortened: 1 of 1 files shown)


dsanders11:
Hi @tkomde, thanks for the PR! The docs in this repo are actually just a mirror from the core repo (https://github.com/electron/electron), so this change will need to be made there. As such I'm going to close this PR, but feel free to open an equivalent PR on that repo.

tkomde:
@dsanders11 Thanks for explaining. I will rebuild this PR over there.

PR #867, opened 2025-09-15
chore: mark 35.0.0 as not supported

ckerr:
#### Description of Change

Mark Electron 35.0.0 as no longer supported in https://www.electronjs.org/docs/latest/tutorial/electron-timelines . 

I'm not sure why `electron-website-docs-updater` didn't catch this?

#### Checklist


dsanders11:
> I'm not sure why `electron-website-docs-updater` didn't catch this?

Our automation is currently busted due to some recent changes (PR status checks now required). I have a PR up which attempts to fix that in #868. I'm going to close this for now so that we have a test case for the changes to our automation script once that PR lands (and this PR would also need to update `docs/latest/.sha` to keep the audit happy).

PR #868, opened 2025-09-17
ci: change automated docs update to use PRs

dsanders11:
#### Description of Change

We have some required status checks now which require a PR, and that's breaking our existing automation solution of pushing straight to `main`. Run everything through a PR which the existing GitHub app will create, and then use the generic GitHub Actions app to approve and merge the PR.

#### Checklist

- [ ] This PR was not created with AI. (PRs created mainly with AI will be closed. They waste our team's time. We ban repeat offenders.)

### DIFF
--- a/.github/workflows/update-docs.yml
+++ b/.github/workflows/update-docs.yml
@@ -18,6 +18,10 @@ jobs:
   update-docs:
     runs-on: ubuntu-latest
     environment: docs-updater
+    env:
+      SHA: ${{ github.event.client_payload.sha || github.event.inputs.sha }}
+    permissions:
+      pull-requests: write
     steps:
       - name: Generate GitHub App token
         uses: electron/github-app-auth-action@384fd19694fe7b6dcc9a684746c6976ad78228ae # v1.1.1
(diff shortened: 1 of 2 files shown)


.github/workflows/update-docs.yml
```diff
@@ -35,14 +39,33 @@ jobs:
...
+            --head "docs/update-to-${SHA}" \
+            --base main
+      - name: Approve and merge pull request
+        shell: bash
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          PR_NUMBER=$(gh pr list --head "docs/update-to-${SHA}" --json number --jq '.[0].number')
+          gh pr review ${PR_NUMBER} --approve
+          gh pr merge ${PR_NUMBER} --auto --squash
```
MarshallOfSound: I think we need a "wait for checks" step here that waits up to ~20 minutes for required CI to pass?

More ideally it would fire a separate job once PR enters "mergable" state. But for now whack a `while !status_passed` kinda loop for fun will get us unblocked I think

dsanders11: I *think* the `--auto` flag should handle this for us? From [the `gh` CLI docs](https://cli.github.com/manual/gh_pr_merge): "Automatically merge only after necessary requirements are met".

.github/workflows/update-docs.yml
```diff
@@ -35,14 +39,32 @@ jobs:
...
+            --title "chore: update ref to docs (🤖)" \
+            --body "Automated PR to update the docs to the latest commit (${SHA})" \
+            --head "docs/update-to-${SHA}" \
+            --base main
+      - name: Approve and merge pull request
+        shell: bash
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh pr review "docs/update-to-${SHA}" --approve
```
erickzhao, line 69: Naive question: would this be a self-approve and then not count for the purposes of mergeability?

dsanders11: There's a subtle difference here (which I could call out better in the code with a comment), but we're using `secrets.GITHUB_TOKEN` here so it'll be the default GitHub Actions app approving the PR, which was opened by our GitHub app.

PR #882, opened 2025-10-07
feat: add markdown output plugin to copy markdown files alongside HTML

clavin:
#### Description of Change

Adds a custom Docusaurus plugin to emit Markdown files alongside the generated HTML files.

When Docusaurus compiles a markdown route, it creates an `index.html` file within that route. This plugin copies the source file to a `index.md` file alongside it.

##### Source tree
```
─ docs/
  ├─ flowers.md
  └─  ...
```

##### Output tree
```
─ docs/
  ├─ flowers/
  │  ├─ index.html
  │  └─ index.md     # New!
  └─ ...
```

#### Checklist

### DIFF
--- a/src/plugins/markdown-output/index.ts
+++ b/src/plugins/markdown-output/index.ts
@@ -1,0 +1,95 @@
+/**
+ * Plugin to output index.md files alongside index.html in the build directory.
+ * This copies the original markdown source files to their corresponding locations
+ * in the build output.
+ */
+
+import fs from 'node:fs/promises';
+import path from 'node:path';
+import { Plugin } from '@docusaurus/types';
+import logger from '@docusaurus/logger';
+
+/** Route prefixes to copy markdown files within. */
+const PREFIXES = ['docs/', 'blog/'];
+
+type Context = Parameters<Plugin['postBuild']>[0];
+
+async function fileExists(filePath: string): Promise<boolean> {
+  return fs.access(filePath).then(
+    () => true,
+    () => false,
+  );
+}
+
+async function findMarkdownSource(
+  routePath: string,
+  { siteDir }: Context,
+): Promise<string | null> {
+  for (const prefix of PREFIXES) {
+    const routePrefix = `/${prefix}`;
+
+    if (!routePath.startsWith(routePrefix)) continue;
+
+    // Since routePrefix is absolute and routePath starts with routePrefix, it too must be absolute
+    // meaning we can safely use path.relative to get the relative path
+    const relativePath = path.relative(routePrefix, routePath);
+    const sourceDir = path.join(siteDir, prefix);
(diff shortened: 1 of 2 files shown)


src/plugins/markdown-output/index.ts
```diff
@@ -0,0 +1,95 @@
...
+const PREFIXES = ['docs/', 'blog/'];
+
+type Context = Parameters<Plugin['postBuild']>[0];
+
+async function fileExists(filePath: string): Promise<boolean> {
+  return fs.access(filePath).then(
+    () => true,
+    () => false,
+  );
+}
```
erickzhao: Do permissions come into play here, or would it be simpler to just use `fs.existsSync`?

clavin: Nope! At least not in a meaningful way. This was just a concise way to do it async.

Would we want to just use [`pathExists` from `fs-extra`](https://github.com/jprichardson/node-fs-extra/blob/HEAD/docs/pathExists.md) instead? Or is the async part the concern? 😅

erickzhao: I'm just curious because we generally use `fs.existsSync` as a replacement for `fs.pathExists` across ecosystem repos nowadays. I think having the `fs.existsSync` call is slightly more readable but also not a big deal either way.

clavin: Was concerned that 370+ sync calls might be slower than async, but I measured it and it turns out to be faster! (_100ms sync / 160ms async_, n = 2.) Turns out the async overhead for these little fs calls is much higher than I expected!

Fixed in bc690b66bb1fb7392285767783c6fc78252b8aec and greatly simplifies the code. Good call! 😄 👍

clavin:
I plan to pair this with a change to send some clients these markdown files: https://github.com/electron/infra/pull/161

src/plugins/markdown-output/index.ts
erickzhao: Woo, thanks for measuring!

erickzhao (approved):

src/plugins/markdown-output/index.ts
```diff
@@ -0,0 +1,83 @@
...
+type Context = Parameters<Plugin['postBuild']>[0];
+
+function findMarkdownSource(
+  routePath: string,
+  { siteDir }: Context,
+): string | null {
+  for (const prefix of PREFIXES) {
+    const routePrefix = `/${prefix}`;
+
+    if (!routePath.startsWith(routePrefix)) continue;
```
dsanders11, line 24: This probably doesn't play nice with i18n as those are prefixed by the language code. I think ideally we'd also serve the translated Markdown, so this should be updated to take that into account.

PR #911, opened 2025-11-03
build: Docusaurus 3.9

erickzhao:
https://docusaurus.io/blog/releases/3.9


dsanders11 (approved):
https://docusaurus.io/blog/releases/3.9#docsearch-v4

We should follow-up at some point by looking into enabling DocSearch v4 and potentially the AskAI feature (looks like it requires some set up on the account side, and don't know that it would be available for free).

Otherwise I did some spot check comparisons between the preview deploy and the site in prod and only visual difference I noticed is the search bar look has been tweaked (which is fine):

| Current Site | Preview Site |
|--------|--------|
| Old | New |

PR #941, opened 2025-11-20
ci: don't fail docs audit if .sha is newer non-docs commit on e/e

dsanders11:
#### Description of Change

Sometimes `docs/latest/.sha` gets out ahead of what we would expect it to be since `electron-website-updater` is supposed to only send any changes to `docs/` on `e/e` (to minimize churn here), but has a bug where it'll send commits that touch any file with the word `docs` in the name. When that happens we can end up ahead of the expected SHA from `e/e`, so be more lenient and as long as we're equal to or ahead of the expected SHA, consider that as pass.

#### Checklist

### DIFF
--- a/.github/workflows/audit-docs-version.yml
+++ b/.github/workflows/audit-docs-version.yml
@@ -69,13 +69,13 @@ jobs:
               if (comparison.status !== 200) {
                 throw new Error('Failed to compare commits');
               }
-        
+
               if (comparison.data.behind_by === 0 && comparison.data.ahead_by >= 0) {
                 core.summary.addRaw('🎉 Docs are up-to-date');
               } else {
                 console.log(`Got ${latestDocsSHA}, expected ${commit.sha}`);
                 core.summary.addRaw('🚨 Docs are NOT up-to-date');
-  
+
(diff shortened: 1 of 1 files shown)


.github/workflows/audit-docs-version.yml
```diff
@@ -58,11 +58,27 @@ jobs:
             }
 
             if (commit.sha !== latestDocsSHA) {
-              console.log(`Got ${latestDocsSHA}, expected ${commit.sha}`);
-              core.summary.addRaw('🚨 Docs are NOT up-to-date');
+              // A bug in `electron-website-updater` sometimes sends up new commits that
+              // touched any file with `docs` in the name, so allow commits to be newer
```
erickzhao, line 62: This is pretty funny and we should fix that

erickzhao (approved):

.github/workflows/audit-docs-version.yml
dsanders11: I have a loose plan in the back of my head to retire `electron-website-updater` and just send the updates from a workflow on `e/e`, but haven't quite gotten to it yet. 😅

PR #958, opened 2025-12-19
blog: add first tech talk blog post

nikwen:
#### Description of Change

Adds our first Tech Talk blog post! :tada:

#### Checklist

### DIFF
--- a/blog/tech-talk-window-resize-behavior.md
+++ b/blog/tech-talk-window-resize-behavior.md
@@ -1,5 +1,5 @@
 ---
-title: 'Tech Talk #1: Improving Window Resize Behavior'
+title: 'Tech Talk: Improving Window Resize Behavior'
 date: 2025-12-19T00:00:00.000Z
 authors:
   - name: nikwen


blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+_We're launching a new blog post series where we share glimpses into our work on Electron. If you find this work interesting, please consider [contributing](https://github.com/electron/electron/)!_
+
+---
+
+Recently, I worked on improving Electron and Chromium's window resize behavior.
+
+## The bug
+
+We were seeing an issue on Windows where old frames would become visible while resizing a window:
```
erickzhao, line 20: Should we link to https://github.com/electron/electron/issues/36280 (and also close that issue?)

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+## What made this bug particularly interesting?
+
+1. It was challenging.
+2. As you'll see later, there were two different bugs under the hood.
+
+## Fixing the bug
+
+With a bug like this, the first challenge is figuring out where to start looking.
+
+Electron builds upon Chromium. Chromium is the open source version of Google Chrome. When compiling Electron, Electron's source code is added into the Chromium source tree as a subdirectory. Electron then relies on Chromium's code to provide most of the functionality of a modern browser.
```
erickzhao: nit: joining two clauses together to flow a bit better

```suggestion
Electron builds upon Chromium, the open source version of Google Chrome. When compiling Electron, Electron's source code is added into the Chromium source tree as a subdirectory. Electron then relies on Chromium's code to provide most of the functionality of a modern browser.
```

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+Additionally, the issue was not visible on macOS. That suggested that it was in Windows-specific source code.
+
+## The crucial lead
+
+I tried a lot of different command line flags and configuration options.
+
+I noticed that [`app.disableHardwareAcceleration()`](../docs/latest/api/app#appdisablehardwareacceleration) fixed the issue. Without hardware acceleration, the issue was gone.
+
+Here is some context: Chromium supports various different graphics APIs for showing pixels on screen (OpenGL, Vulkan, Metal, and more). On Windows, it uses different graphics APIs than on Windows or Linux. Even on Windows, Chromium can work with multiple different graphics backends.
```
erickzhao:
```suggestion
Here is some context: Chromium supports various different graphics APIs for showing pixels on screen (OpenGL, Vulkan, Metal, and more). On Windows, it uses different graphics APIs than on macOS or Linux. Even on Windows, Chromium can work with multiple different graphics backends.
```

typo I think

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+- `--use-gl=egl`
+- `--use-gl=osmesa`
+- `--use-gl=swiftshader`
+
+The following flags reproduced the issue:
+
+- `--use-angle=d3d11` (this is currently the default on Windows)
+- `--use-angle=gl` (falls back to Direct3D 11 on Windows, see `chrome://gpu/`)
+
+None of the working flags were good enough to be used as the default in Electron apps on Windows. They were either too slow or lacked good driver support.
```
erickzhao:
```suggestion
None of the working flags were suitable as the default backend for Electron apps on Windows. They were either too slow or lacked good driver support.
```

Small edit to avoid repeating `good` in consecutive sentences

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+Which graphics backend Chromium uses depends on the user's hardware. For example, some graphics backends require the computer to have a GPU.
+
+I tried various graphics backends and noticed that the following flags fixed the issue:
+
+- `--use-angle=warp`
+- `--use-angle=vulkan`
+- `--use-gl=desktop`
+- `--use-gl=egl`
+- `--use-gl=osmesa`
+- `--use-gl=swiftshader`
```
erickzhao, line 69: In the narrative, I think you could explain what the difference between calling `--use-angle` and `--use-gl` represents for Chromium and avoid needing to explain ANGLE at the bottom after we've already name-dropped it a few times.

As an aside, TIL that ANGLE stands for "Almost Native Graphics Layer Engine". Funny name!

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+By chance, I stumbled across a few helpful debugging flags in the Chromium source code:
+
+- `--ui-show-paint-rects`
+- `--ui-show-property-changed-rects`
+- `--ui-show-surface-damage-rects`
+- `--ui-show-composited-layer-borders`
+- `--tint-composited-content`
+- `--tint-composited-content-modulate`
+- (And more)
```
erickzhao, line 96: Is there a specific file we want to reference here for folks who like to click through and follow along? (I know this is in the "less technical" section but it'd be nice since you explicitly mention the source code here.)

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+That allowed me to see which part of the graphics stack was producing which output.
+
+In particular, the combination of `--tint-composited-content` and `--tint-composited-content-modulate` was really helpful. The former adds a tint to the output of the compositor. The latter changes the tint color on every frame.
+
+![Screenshot of Chromium with the --tint-composited-content flag](/assets/img/blog/tech-talk-window-resize-behavior/tinting-composited-content.png)
+
+In the screenshot, the cyan-tinted frame was the last frame that was being drawn.
+
+The jank to the right of that frame was not tinted cyan. It was tinted in different colors that were still there from previous frames. This indicated that the jank was not coming from the compositor. The compositor was sending the right output.
```
erickzhao, line 108: Combining a few more clauses:

```suggestion
The jank to the right of that frame was not tinted cyan, but rather in different colors that were still there from previous frames. This confirmed that the jank was not coming from the compositor, which was was sending the right output.
```

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+The compositor is part of Chromium's graphics stack. The following is very simplified, but for the purpose of this blog post you can imagine it like this:
+
+1. The compositor `cc` produces a `CompositorFrame`, which contains draw instructions.
+2. `cc` sends that `CompositorFrame` to the display compositor `viz`.
+3. `viz` then draws the frame and shows it on screen.
+
+Tinting each `CompositorFrame` showed that the compositor produced the right output. So the issue had to be in the display compositor `viz`.
```
erickzhao, line 116: I think moving this paragraph closer to Line 97 where you first introduce the graphics stack would make it clearer what the compositor does before talking about its output (i.e. it's clearer if the audience understands that `cc` -> `viz` before we mention that the compositor has correct output).

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+## Locating the relevant `viz` code
+
+From there, I started searching for mentions of Direct3D in the `viz` source code.
+
+_Note: From here on, the post will get a bit more technical and reference source code symbols._
```
erickzhao, line 122:
```suggestion
> [!NOTE]
> From here on, the post will get a bit more technical and reference source code symbols.
```

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+I found that on the ANGLE Direct3D 11 backend, Chromium uses the Windows [DirectComposition](https://learn.microsoft.com/en-us/windows/win32/directcomp/directcomposition-portal) API for drawing the window contents.
+
+DirectComposition differs from all other output devices in Chromium. The DirectComposition `OutputSurface` has the capability `supports_viewporter` ([source link 1](https://source.chromium.org/chromium/chromium/src/+/main:ui/gl/dcomp_presenter.cc;l=144-146;drc=1cffe9643e1e70f1f369bbb28e015d9c5f968546), [source link 2](https://source.chromium.org/chromium/chromium/src/+/main:components/viz/service/display/output_surface.h;l=100-103;drc=c96a878b1cb45f60aac2285ffbdbd6b53dc92415)).
```
erickzhao: > DirectComposition differs from all other output devices in Chromium

nit: should we explain what "output device" means in this context? Without context, this makes me think of audio/video input/output devices.

erickzhao:
```suggestion
Unlike other output devices in Chromium, the DirectComposition `OutputSurface` has the `supports_viewporter` capability ([source link 1](https://source.chromium.org/chromium/chromium/src/+/main:ui/gl/dcomp_presenter.cc;l=144-146;drc=1cffe9643e1e70f1f369bbb28e015d9c5f968546), [source link 2](https://source.chromium.org/chromium/chromium/src/+/main:components/viz/service/display/output_surface.h;l=100-103;drc=c96a878b1cb45f60aac2285ffbdbd6b53dc92415)).
```

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+An output surface is a bitmap that can be drawn to, often backed by a GPU texture.
+
+Without `supports_viewporter`, whenever the window size changes, Chromium will create a new output surface matching the new window size. Then it will draw on that surface and show it.
```
erickzhao, line 130:
```suggestion
Without `supports_viewporter`, Chromium will create a new output surface matching the new window size whenever the window size changes. Then it will draw on that surface and show it.
```

Shifting the clause here to make the sentence roll a bit smoother

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+`supports_viewporter` tries to reduce these costly surface allocations. With `supports_viewporter`, Chromium will not allocate a new surface on every resize. Instead, it will allocate a surface that is too large for what we need to draw. Then it will only paint to and show a certain sub-rectangle (the "viewport") of that surface on screen. The other parts of the surface are not supposed to be shown on screen.
+
+This is supposed to make resizing more efficient because all Chromium needs to do is pad the surface to the proper width and height instead of allocating a new surface on every resize. This surface resize logic lives [here](https://source.chromium.org/chromium/chromium/src/+/main:components/viz/service/display/direct_renderer.cc;l=1082;drc=c96a878b1cb45f60aac2285ffbdbd6b53dc92415).
```
erickzhao:
```suggestion
This is supposed to make resizing more efficient because all Chromium needs to do is pad the surface to the proper width and height instead of allocating a new surface on every resize. This surface resize logic lives in [`direct_renderer.cc`](https://source.chromium.org/chromium/chromium/src/+/main:components/viz/service/display/direct_renderer.cc;l=1082;drc=c96a878b1cb45f60aac2285ffbdbd6b53dc92415).
```

prefer descriptive markdown link titles

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+Here's what that looks like:
+
+![Visualization showing the surface, viewport, and clip rect](/assets/img/blog/tech-talk-window-resize-behavior/visualization-1.png)
+
+The blue rectangle is our surface. The green area is our viewport, i.e., the area of our surface that is supposed to be visible and that we actively draw to. The red rectangle is our clip rect(angle), i.e., the part of the surface that is actually being shown on screen.
```
erickzhao: Could move this to a bullet list to make it read a bit easier.

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+As a performance optimization, only the viewport (the green area) is repainted when we get a new frame. The rest is left unchanged. This is important. We only ever repaint the green viewport. We don't update the areas outside of the viewport.
+
+When we resize the window, what's supposed to happen is that in an atomic transaction (= at the exact same time) we repaint the viewport (= the area that's supposed to be visible on screen) and then update the clip rect to clip the surface to the new viewport size.
```
erickzhao, line 148:
```suggestion
When we resize the window, what's supposed to happen is that in an atomic transaction (i.e. at the exact same time) we repaint the viewport (i.e. the area that's supposed to be visible on screen) and then update the clip rect to clip the surface to the new viewport size.
```

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+After the resize, it should look like this:
+
+![Visualization with updated viewport and clip rect](/assets/img/blog/tech-talk-window-resize-behavior/visualization-2.png)
+
+And that's where we get to the first of our two bugs.
+
+## First bug
+
+Sometimes these operations can get out of sync. For example, the clip rect might get updated before the viewport is repainted. Then we get a result like this:
```
erickzhao, line 158:
```suggestion
Sometimes these operations can get out of sync. For example, if clip rect gets updated before the viewport is repainted, we get a result like this:
```

Combining a few sentences for flow

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+![Visualization where the clip rect was updated before the viewport](/assets/img/blog/tech-talk-window-resize-behavior/visualization-3.png)
+
+We still show the old frame in the green viewport. But the clip rect has become larger and we show areas of the surface that we haven't repainted yet.
```
erickzhao, line 162:
```suggestion
We still show the old frame in the green viewport, but the clip rect has become larger and we show areas of the surface that we haven't repainted yet.
```

`but` makes it a dependent clause

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+Then parts of the clip rect would still show the previous frame because the new frame was smaller and we did not repaint any areas beyond the new viewport.
+
+Now why do these operations not happen in sync?
+
+We use two different Windows APIs here:
+
+- [`IDXGISwapChain1::Present1`](https://source.chromium.org/chromium/chromium/src/+/main:gpu/command_buffer/service/shared_image/dxgi_swap_chain_image_backing.cc;l=283;drc=92e802d8fa66bafb0cca9ed32143d6148d8e0411) — This shows the new pixels on screen / updates the new viewport.
+- [`IDCompositionDevice::Commit`](https://source.chromium.org/chromium/chromium/src/+/main:ui/gl/dc_layer_tree.cc;l=1034;drc=35d26d364efb57d0386b98312ba739f7f65ae97e) — This updates the clip rect.
+
+Here's what's important to understand: both functions return synchronously on the CPU. However, they schedule tasks that run asynchronously on the GPU at a later time. Windows and its services (such as [DWM](https://en.wikipedia.org/wiki/Desktop_Window_Manager)) decide when they will run and in which order. So they take effect asynchronously, and not always within the same frame.
```
erickzhao:
```suggestion
Here's what's important to understand: both functions return synchronously on the CPU. However, they schedule tasks that run asynchronously on the GPU at a later time. Windows and its services (such as [DWM](https://en.wikipedia.org/wiki/Desktop_Window_Manager)) decide when these tasks will run and in which order. So they take effect asynchronously, and not always within the same frame.
```

`they` is ambiguous here since we mention both "functions" and "tasks" in the previous few sentences. Did you mean the async tasks here?

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+Unfortunately, Windows provides no way for us to synchronize those operations. So I had to find other approaches to fix this.
```
erickzhao, line 181:
```suggestion
Unfortunately, Windows provides no way for us to synchronize those operations, so I had to find other approaches to fix this.
```

`so` indicates a dependent clause

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+2. While resizing, switch from an `IDXGISwapChain1` to a DirectComposition surface which synchronizes updates with `IDCompositionDevice::Commit`. This also fixes the artifacts.
+
+We went with the first option because it leads to faster resizes than the second option.
+
+I landed a [patch](https://crrev.com/c/7115438) in Chromium that implemented that first solution.
+
+I also submitted two other patches in preparation for the main patch:
+
+1. The [first one](https://crrev.com/c/7129658) fixed a bug in existing code that would make CI fail in combination with the main patch. It also made launching Electron apps and Chrome a tiny bit faster.
+2. The [second one](https://crrev.com/c/7210913) was split off to make code review easier on the main patch.
```
erickzhao, line 195: nit: can we also make these link texts like `crrev.com/c/7156576` below?

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+When the user resizes the window, Chromium needs to redraw the contents of the window for the new window size. This takes some time. The new frame isn't ready immediately.
+
+Here's a sequence of frames that demonstrates this:
+
+![First frame of the Chromium resize sequence](/assets/img/blog/tech-talk-window-resize-behavior/chromium-resize-sequence-1.png)
+![Second frame of the Chromium resize sequence](/assets/img/blog/tech-talk-window-resize-behavior/chromium-resize-sequence-2.png)
+![Third frame of the Chromium resize sequence](/assets/img/blog/tech-talk-window-resize-behavior/chromium-resize-sequence-3.png)
+![Fourth frame of the Chromium resize sequence](/assets/img/blog/tech-talk-window-resize-behavior/chromium-resize-sequence-4.png)
+![Fifth frame of the Chromium resize sequence](/assets/img/blog/tech-talk-window-resize-behavior/chromium-resize-sequence-5.png)
```
erickzhao, line 211: Should we splice this into a `GIF` with slower timings between the frames?

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+I landed a [patch](https://crrev.com/c/7115438) in Chromium that implemented that first solution.
+
+I also submitted two other patches in preparation for the main patch:
+
+1. The [first one](https://crrev.com/c/7129658) fixed a bug in existing code that would make CI fail in combination with the main patch. It also made launching Electron apps and Chrome a tiny bit faster.
+2. The [second one](https://crrev.com/c/7210913) was split off to make code review easier on the main patch.
+
+## Second bug
+
+In addition to this first bug, there was a second bug which also led to stale pixels.
```
erickzhao, line 199:
```suggestion
In addition to the `DirectComposition` issue, I discovered a second bug that also led to stale pixels.
```

Active voice!

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+Here's what was going on there:
+
+When the user resizes the window, Chromium needs to redraw the contents of the window for the new window size. This takes some time. The new frame isn't ready immediately.
```
erickzhao, line 203:
```suggestion
When the user resizes the window, Chromium needs to redraw the contents of the window for the new window size. This takes some time, and the new frame isn't ready immediately.
```

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+Here's a sequence of frames that demonstrates this:
+
+![First frame of the Chromium resize sequence](/assets/img/blog/tech-talk-window-resize-behavior/chromium-resize-sequence-1.png)
+![Second frame of the Chromium resize sequence](/assets/img/blog/tech-talk-window-resize-behavior/chromium-resize-sequence-2.png)
+![Third frame of the Chromium resize sequence](/assets/img/blog/tech-talk-window-resize-behavior/chromium-resize-sequence-3.png)
+![Fourth frame of the Chromium resize sequence](/assets/img/blog/tech-talk-window-resize-behavior/chromium-resize-sequence-4.png)
+![Fifth frame of the Chromium resize sequence](/assets/img/blog/tech-talk-window-resize-behavior/chromium-resize-sequence-5.png)
+
+At a certain time during the resize, Windows tells us: "The window is 1,000 pixels wide." But the frames that the browser compositor produces are lagging behind. The last frame that we have painted might be 600 pixels wide.
```
erickzhao, line 213:
```suggestion
At a certain time during the resize, Windows tells us: "The window is 1,000 pixels wide." However, the frames that the browser compositor produces are lagging behind: the last frame that we have painted might be 600 pixels wide.
```

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+Historically, Chromium used to skip frames where the width of the window did not match the width of the frame that it last painted. It would decide to just not update the window.
+
+However, that would often lead to the window contents not being updated at all until the resize operation was finished.
+
+So [in 2015](https://crrev.com/1426453006) someone decided: "Why not show these frames? They might not match the window size perfectly, but at least we can show _something_."
+
+It would lead to gutter, but at that time the gutter was black. So that was better than the previous implementation.
```
erickzhao, line 221:
```suggestion
It would lead to gutter, but at that time the gutter was black. So that was better than the previous implementation.
```

"lead to gutter": I think maybe mentioning the gutter at the first sentence of this section might be a good lead-in to explain that it was this bug but the stale pixels were black.

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+Now, 10 years later with DirectComposition, that gutter was often filled with stale pixels.
+
+Let's look at what was happening there:
+
+Every frame consists of multiple render passes. The render passes represent the various things that should be drawn on screen. From complicated bitmaps to rectangles filled with solid colors.
```
erickzhao:
```suggestion
Every frame consists of multiple render passes. These render passes represent the various things that should be drawn on screen, from complicated bitmaps to rectangles filled with solid colors.
```

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+Every frame has a root render pass, which contains all other render passes and glues them together. (Render passes are arranged in a tree structure and the root render pass is the root of that tree.)
+
+So now during a resize, we'd get to a point where we know the window is 1,000 pixels wide. Accordingly, we'd adjust the viewport of our output surface to also be 1,000 pixels wide. But the frame that we just received is only 600 pixels wide.
```
erickzhao, line 231:
```suggestion
During a resize, we'd get to a point where we know the window is 1,000 pixels wide. Accordingly, we'd adjust the viewport of our output surface to also be 1,000 pixels wide, but the frame that we just received is only 600 pixels wide.
```

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+The [optimization from 2015](https://source.chromium.org/chromium/chromium/src/+/main:components/viz/service/display/display.cc;l=1053-1062;drc=152a83bb9fcba4ea3a7b9f8efbe0f6204f95eadd) would then go and change the width of the root render pass to also be 1,000 pixels. But it wouldn't change what the render passes would actually draw on screen. They'd still only contain instructions to draw a picture that is 600 pixels wide.
```
erickzhao, line 233:
```suggestion
The [optimization from 2015](https://source.chromium.org/chromium/chromium/src/+/main:components/viz/service/display/display.cc;l=1053-1062;drc=152a83bb9fcba4ea3a7b9f8efbe0f6204f95eadd) would then go and change the width of the root render pass to also be 1,000 pixels, but wouldn't change what the render passes would actually draw on screen. They'd still only contain instructions to draw a picture that is 600 pixels wide.
```

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+I landed a fix for this issue in [crrev.com/c/7156576](https://crrev.com/c/7156576).
+
+The fix changes what we do when we receive a frame with a different size than our window. Instead of resizing the frame and adding gutter that contains stale pixels, we [resize our viewport](https://source.chromium.org/chromium/chromium/src/+/main:components/viz/service/frame_sinks/root_compositor_frame_sink_impl.cc;l=555-556;drc=78bd63326c0a2489e6b016140aa4d84014af077a) and our clip rect.
+
+![Visualization where the clip rect and viewport size are adjusted to the frame size](/assets/img/blog/tech-talk-window-resize-behavior/visualization-6.png)
+
+We clip our surface to the size of the frame that we received. We don't show anything beyond the 600 pixels that we have draw instructions for.
+
+Voila, no more gutter, no more stale pixels!
```
erickzhao:
```suggestion
Voilà, no more gutter, no more stale pixels!
```

the francophone in me wants to put the correct accents

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+(Side note: Without `supports_viewporter`, this would be an expensive operation because it would allocate a new output surface. However, with DirectComposition, we use the "viewporter" feature. So we don't reallocate the surface when we change the viewport size. We just make a different portion of it visible. Thus, it is a cheap operation.)
```
erickzhao:
```suggestion
> [!NOTE]
> Without `supports_viewporter`, this would be an expensive operation because it would allocate a new output surface. However, with DirectComposition, we use the "viewporter" feature to avoid reallocating the surface when we change the viewport size. We just make a different portion of it visible. Thus, it is a cheap operation.)
```

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+## Backporting the patches to Electron
+
+Once the fixes made it into Chromium, we had to pull them into Electron, too.
```
erickzhao, line 264: Would be good to link to https://www.electronjs.org/docs/latest/tutorial/electron-versioning somewhere maybe

blog/tech-talk-window-resize-behavior.md
nikwen: 1. I decided not to link to it because there are two different issues being reported in that issue and that's confusing:
  1. The resize being slow (not fixed).
  2. Old pixels showing (fixed).
2. For that reason, I also wouldn't close it.

blog/tech-talk-window-resize-behavior.md
nikwen: Thanks for catching this!

blog/tech-talk-window-resize-behavior.md
nikwen: Unfortunately, they are all in different files, which makes it hard to link to a specific file.

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+## The bug
+
+We were seeing an issue on Windows where old frames would become visible while resizing a window:
+
+![Animated GIF showing the issue where old frames would be shown while resizing windows](/assets/img/blog/tech-talk-window-resize-behavior/window-resize-bug-showing-old-frames.gif)
+
+## What made this bug particularly interesting?
+
+1. It was challenging.
```
mitchchn, line 26: Expand on this a bit? e.g., it was challenging because it involved going deep into GPU timing code in Chromium.

blog/tech-talk-window-resize-behavior.md
mitchchn, line 20: I'd explicitly add something like ", causing tearing and jankiness" to drive home *why* it's a problem to show old frames.

blog/tech-talk-window-resize-behavior.md
nikwen: Thanks for the suggestion! I'd prefer not to change this. I prefer shorter sentences.

blog/tech-talk-window-resize-behavior.md
nikwen: Thanks!

blog/tech-talk-window-resize-behavior.md
nikwen: Thanks for the suggestion! I'd prefer not to change this. I prefer shorter sentences.

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+Additionally, the issue was not visible on macOS. That suggested that it was in Windows-specific source code.
+
+## The crucial lead
+
+I tried a lot of different command line flags and configuration options.
+
+I noticed that [`app.disableHardwareAcceleration()`](../docs/latest/api/app#appdisablehardwareacceleration) fixed the issue. Without hardware acceleration, the issue was gone.
+
+Here is some context: Chromium supports various different graphics APIs for showing pixels on screen (OpenGL, Vulkan, Metal, and more). On Windows, it uses different graphics APIs than on macOS or Linux. Even on Windows, Chromium can work with multiple different graphics backends.
```
mitchchn, line 58:
```suggestion
Here is some context: Chromium supports various graphics APIs for showing pixels on screen (OpenGL, Vulkan, Metal, and more). On Windows, Chromium can work with multiple different graphics backends.
```

Trimmed out redundant line and tightened up.

blog/tech-talk-window-resize-behavior.md
nikwen: Good catch. Thanks!

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+## Backporting the patches to Electron
+
+Once the fixes made it into Chromium, we had to pull them into Electron, too.
+
+On the `main` branch, Electron updates its Chromium version constantly. As a result, the patches were merged into `main` in a [Chromium roll PR](https://github.com/electron/electron/pull/49145).
+
+However, commits that make it into `main` right now will only be included in an Electron release in about three months. Our existing release and pre-release branches run on older Chromium versions.
+
+The Chromium patches made it into Chromium 145. Electron 39 runs on Chromium 142. The upcoming Electron 40 will run on Chromium 144. (For comparison, the stable channel of Google Chrome is based on Chromium 143 right now.)
```
mitchchn: All this is true but disorienting for non-maintainers to keep track of. I think your previous paragraph explains well enough why the patches need to be backported without detailing how all the versions line up.

```suggestion
```

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+Big thanks to [Plasticity](https://www.plasticity.xyz) for funding this work!
+
+Thanks to Michael Tang and Vasiliy Telezhnikov from the Chromium team for their help.
+
+## Final thoughts
+
+This was the hardest bug I have ever worked on (and I have worked on many hard bugs in 18 years of writing software).
+
+But it was also one of the most fun projects I have ever worked on.
+
```
mitchchn, line 293: Maybe a hot take but I think there's room here for a sentence or two to **explicitly** call out a misconception people have before going into the CTA: A lot of people believe that Electron doesn't care about performance or native platform quality, but the truth is the exact opposite: we care deeply and we go above and beyond to improve fluidity even in edge cases like this one. And by fixing these bugs, we're improving the experience not just for Electron apps, but also for Chrome and every browser and app that's based on it.

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+Thus, the next step was to backport the patches to [Electron 39](https://github.com/electron/electron/pull/49138) and [Electron 40](https://github.com/electron/electron/pull/49191).
+
+Electron keeps a list of Chromium patches in the [`patches/chromium` directory](https://github.com/electron/electron/tree/a90ccc753b4d38266323dd054db2b98a45917117/patches/chromium). When we backport a Chromium patch, we add it there. When building Electron, these patches are applied to the Chromium source code.
+
+(In general, we try to keep the number of Chromium patches low. Every patch can lead to merge conflicts during Chromium updates. The maintenance burden from patches is real.)
+
+The Electron 39 [backport PR](https://github.com/electron/electron/pull/49138) was merged pretty quickly. The fix became part of [Electron 39.2.6](https://releases.electronjs.org/release/v39.2.6). 🎉
+
+If you resize a window on Electron 39.2.6 or later, you'll see no more stale pixels.
```
mitchchn, line 278: Do you have a gif of smooth resizing in this article? If so it should at least be shown here, but possibly also near the top as a before/after comparison.

blog/tech-talk-window-resize-behavior.md
nikwen: Thanks for the suggestion! My take is that this puts way too much attention on something that doesn't deserve attention and takes attention away from the actually interesting lines around it:

image

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+When we resize the window, what's supposed to happen is that in an atomic transaction (= at the exact same time) we repaint the viewport (= the area that's supposed to be visible on screen) and then update the clip rect to clip the surface to the new viewport size.
+
+After the resize, it should look like this:
+
+![Visualization with updated viewport and clip rect](/assets/img/blog/tech-talk-window-resize-behavior/visualization-2.png)
+
+And that's where we get to the first of our two bugs.
+
+## First bug
```
mitchchn, line 156: The article would be easier to skim if the heading included some detail: "Bug #1: frame painting operations were out of sync"

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+We went with the first option because it leads to faster resizes than the second option.
+
+I landed a [patch](https://crrev.com/c/7115438) in Chromium that implemented that first solution.
+
+I also submitted two other patches in preparation for the main patch:
+
+1. The [first one](https://crrev.com/c/7129658) fixed a bug in existing code that would make CI fail in combination with the main patch. It also made launching Electron apps and Chrome a tiny bit faster.
+2. The [second one](https://crrev.com/c/7210913) was split off to make code review easier on the main patch.
+
+## Second bug
```
mitchchn, line 197: Something like: "Bug #2: Chromium was not redrawing the entire viewport"

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+We use two different Windows APIs here:
+
+- [`IDXGISwapChain1::Present1`](https://source.chromium.org/chromium/chromium/src/+/main:gpu/command_buffer/service/shared_image/dxgi_swap_chain_image_backing.cc;l=283;drc=92e802d8fa66bafb0cca9ed32143d6148d8e0411) — This shows the new pixels on screen / updates the new viewport.
+- [`IDCompositionDevice::Commit`](https://source.chromium.org/chromium/chromium/src/+/main:ui/gl/dc_layer_tree.cc;l=1034;drc=35d26d364efb57d0386b98312ba739f7f65ae97e) — This updates the clip rect.
+
+Here's what's important to understand: both functions return synchronously on the CPU. However, they schedule tasks that run asynchronously on the GPU at a later time. Windows and its services (such as [DWM](https://en.wikipedia.org/wiki/Desktop_Window_Manager)) decide when these tasks will run and in which order. So they take effect asynchronously, and not always within the same frame.
+
+Unfortunately, Windows provides no way for us to synchronize those operations. So I had to find other approaches to fix this.
+
+There were two options that I evaluated with the Chromium maintainers:
```
mitchchn, line 183: Trimming out the discussion about different possible solutions and focusing on the one you implemented would be a good way to shorten the article and tighten it up.

blog/tech-talk-window-resize-behavior.md
nikwen: Done. Thanks!

mitchchn: Feels like this side note can be completely omitted — if someone asks about performance on social media you can always jump in and clarify.

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+Let's look at what was happening there:
+
+Every frame consists of multiple render passes. The render passes represent the various things that should be drawn on screen. From complicated bitmaps to rectangles filled with solid colors.
+
+Every frame has a root render pass, which contains all other render passes and glues them together. (Render passes are arranged in a tree structure and the root render pass is the root of that tree.)
+
+So now during a resize, we'd get to a point where we know the window is 1,000 pixels wide. Accordingly, we'd adjust the viewport of our output surface to also be 1,000 pixels wide. But the frame that we just received is only 600 pixels wide.
+
+The [optimization from 2015](https://source.chromium.org/chromium/chromium/src/+/main:components/viz/service/display/display.cc;l=1053-1062;drc=152a83bb9fcba4ea3a7b9f8efbe0f6204f95eadd) would then go and change the width of the root render pass to also be 1,000 pixels. But it wouldn't change what the render passes would actually draw on screen. They'd still only contain instructions to draw a picture that is 600 pixels wide.
```
mitchchn, line 233: I think the discussion about the 2015 optimization could be tightened up. It's kind of spread throughout this section and repeated multiple times, but even with all the detail it's not entirely clear to me why the optimization was appropriate in 2015 but causes janky resizing today.

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+The Chromium patches made it into Chromium 145. Electron 39 runs on Chromium 142. The upcoming Electron 40 will run on Chromium 144. (For comparison, the stable channel of Google Chrome is based on Chromium 143 right now.)
+
+Thus, the next step was to backport the patches to [Electron 39](https://github.com/electron/electron/pull/49138) and [Electron 40](https://github.com/electron/electron/pull/49191).
+
+Electron keeps a list of Chromium patches in the [`patches/chromium` directory](https://github.com/electron/electron/tree/a90ccc753b4d38266323dd054db2b98a45917117/patches/chromium). When we backport a Chromium patch, we add it there. When building Electron, these patches are applied to the Chromium source code.
+
+(In general, we try to keep the number of Chromium patches low. Every patch can lead to merge conflicts during Chromium updates. The maintenance burden from patches is real.)
+
+The Electron 39 [backport PR](https://github.com/electron/electron/pull/49138) was merged pretty quickly. The fix became part of [Electron 39.2.6](https://releases.electronjs.org/release/v39.2.6). 🎉
```
mitchchn, line 276: It's a bit repetitive to link the backport PR again and repeat the version number several times. We can merge this with the next paragraph:

```suggestion
The patches were merged pretty quickly, and if you resize a window in [Electron 39.2.6](https://releases.electronjs.org/release/v39.2.6) or later, you'll see no more stale pixels. 🎉
```

blog/tech-talk-window-resize-behavior.md
nikwen: Added. Thanks!

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+On the `main` branch, Electron updates its Chromium version constantly. As a result, the patches were merged into `main` in a [Chromium roll PR](https://github.com/electron/electron/pull/49145).
+
+However, commits that make it into `main` right now will only be included in an Electron release in about three months. Our existing release and pre-release branches run on older Chromium versions.
+
+The Chromium patches made it into Chromium 145. Electron 39 runs on Chromium 142. The upcoming Electron 40 will run on Chromium 144. (For comparison, the stable channel of Google Chrome is based on Chromium 143 right now.)
+
+Thus, the next step was to backport the patches to [Electron 39](https://github.com/electron/electron/pull/49138) and [Electron 40](https://github.com/electron/electron/pull/49191).
+
+Electron keeps a list of Chromium patches in the [`patches/chromium` directory](https://github.com/electron/electron/tree/a90ccc753b4d38266323dd054db2b98a45917117/patches/chromium). When we backport a Chromium patch, we add it there. When building Electron, these patches are applied to the Chromium source code.
```
mitchchn, line 272: Again in the interest of tightening up the post, I think it's a bit too literal/explicit to talk about the patch directory and state that patches are applied during build. We can probably do without this paragraph completely — if people click on the PR link, they can see the structure right there. We can also link to the docs about Chromium patches in the next paragraph without stating this information.

```suggestion
```

blog/tech-talk-window-resize-behavior.md
nikwen: Fixed in a slightly different way. Thanks!

blog/tech-talk-window-resize-behavior.md
```diff
@@ -0,0 +1,290 @@
...
+
+(In general, we try to keep the number of Chromium patches low. Every patch can lead to merge conflicts during Chromium updates. The maintenance burden from patches is real.)
```
mitchchn: This would be a good place to link directly to the docs and also clarify that the patch is both justified and temporary.

```suggestion
In general, we [try to keep the number of Chromium patches low](https://www.electronjs.org/docs/latest/development/patches). Patches lead to merge conflicts during Chromium updates, and the maintenance burden is real.

In this case, the significant performance improvement justified that burden. It also helps to know that the patch is temporary and will soon be addressed upstream.
```

blog/tech-talk-window-resize-behavior.md
nikwen: Tbh I'd prefer to keep it as it. I think it makes it easier to understand.

Here the relationship is "A leads to B". It's easier for the brain to understand in that order than if we mention B first and then bring up the condition for it, A. (See [here](https://web.archive.org/web/20250116041707/https://dilbertblog.typepad.com/the_dilbert_blog/2007/06/the_day_you_bec.html).)

I'd sacrifice smoothness for easier understanding here, but I see where you're coming from.

mitchchn (changes requested):
@nikwen I love this article. I do think it's too long at 3000 words and would read better if it were trimmed down to ~1500. In particular, the sections about the two individual bugs could be made shorter and easier to skim and digest without requiring deep reading.

It's not that you should be afraid to be technical in those sections, but even people who are familiar with Chromium will understand you better if you summarize your findings and changes instead of walking through the complete process you went through when investigating/debugging.

blog/tech-talk-window-resize-behavior.md
nikwen: Good idea. Thanks!

blog/tech-talk-window-resize-behavior.md
nikwen: We could. This is something that I didn't do for time reasons.

blog/tech-talk-window-resize-behavior.md
nikwen: My initial draft while writing it mentioned the `crrev.com` links. I removed them because it really doesn't add any value to the reader if we spell them out.

blog/tech-talk-window-resize-behavior.md
nikwen: Thanks for the suggestion! I tested this and tbh I find it easier to read with the equal signs:

image

vs.

image

blog/tech-talk-window-resize-behavior.md
nikwen: Nevermind, let's get this in. Sounds better. :)

blog/tech-talk-window-resize-behavior.md
nikwen: Good suggestion. Thanks! Added another bullet point.

blog/tech-talk-window-resize-behavior.md
nikwen: Thanks for the suggestion! I thought about this. I ultimately don't think it adds value to the reader to explain the difference in detail. That's why I omitted it.

> TIL that ANGLE stands for "Almost Native Graphics Layer Engine". Funny name!

💯😄

blog/tech-talk-window-resize-behavior.md
nikwen: Good idea. Done. Thanks!

blog/tech-talk-window-resize-behavior.md
nikwen: Thanks! I agree that this could be improved. If you're okay with it, I'd leave it in this order though because I don't have the motivation right now to rephrase it. (Moving it without edits doesn't feel right either.)

blog/tech-talk-window-resize-behavior.md
nikwen: Good point! Given that the rest of the blog post is only about the output surface and doesn't mention the output device at all, let's just remove the mention of the output device.

blog/tech-talk-window-resize-behavior.md
nikwen: Thanks for suggesting. To be honest, I'd prefer to keep this as it is.

blog/tech-talk-window-resize-behavior.md
nikwen: School teachers tell us not to start sentences with "but."

But there's [no actual grammar rule against it](https://prowritingaid.com/starting-a-sentence-with-and-or-but). :)

Sometimes I like it to avoid long sentences. Short sentences are much easier to read.

blog/tech-talk-window-resize-behavior.md
nikwen: Same [as with "but."](https://github.com/electron/website/pull/958#discussion_r2636493237)

blog/tech-talk-window-resize-behavior.md
nikwen: Thanks for the suggestion! To be honest, I'd prefer to keep it as is. In general, I agree that active voice is better. Here I prefer the more neutral phrasing instead of highlighting my achievements (as in "I did this").

blog/tech-talk-window-resize-behavior.md
nikwen: Thanks for suggesting! I'd prefer to keep this as is. I often think it's easier to read many short sentences instead of chaining them together.

blog/tech-talk-window-resize-behavior.md
nikwen: Thanks for the suggestion! If you don't mind, I'd leave this as is.

blog/tech-talk-window-resize-behavior.md
nikwen: Thanks for suggesting! What do you mean by that?

blog/tech-talk-window-resize-behavior.md
nikwen: Thanks! Changed "The" -> "These". With the "from" part, I'm bending the rules of grammar a bit. But I find it much more readable to separate it with a dot than with a comma.

blog/tech-talk-window-resize-behavior.md
nikwen: Thanks! I appreciate the suggestion. Tbh I like it better as is.

blog/tech-talk-window-resize-behavior.md
nikwen: School teachers tell us not to start sentences with "but."

But there's [no actual grammar rule against it](https://prowritingaid.com/starting-a-sentence-with-and-or-but). :)

Sometimes I like it to avoid long sentences. Short sentences are much easier to read.

blog/tech-talk-window-resize-behavior.md
nikwen: Thanks for suggesting! Initially, I had much longer text there.

I cut it because 1) I thought the image did a good job communicating that it was ugly and 2) because I wanted to hook the viewer faster with the next paragraph "why this is interesting".

I'd prefer to keep it as is.

blog/tech-talk-window-resize-behavior.md
nikwen: Thanks for suggesting! I'd prefer to keep it as is.

I think it's interesting to mention that even on the same OS Chromium can use 5 different graphics APIs.

blog/tech-talk-window-resize-behavior.md
nikwen: Thanks! Good point. We can get rid of that.

blog/tech-talk-window-resize-behavior.md
nikwen: Thanks for the suggestion!

I thought about what to include in this section, and I wrote it for someone who is curious about contributing to Electron so that they learn more about how the project works.

When codebytere first told me about the patches directory, I found it incredibly fascinating.

Thus, I'd like to keep it in. :)

blog/tech-talk-window-resize-behavior.md
nikwen: Thanks!

I have added the link.

> In this case, the significant performance improvement justified that burden. It also helps to know that the patch is temporary and will soon be addressed upstream.

That's not fully true. This patch is not a maintenance burden because it's only on the `39-x-y` and `40-x-y` branches. We don't need it on `main` because `main` has the patch already through a Chromium upgrade.

blog/tech-talk-window-resize-behavior.md
nikwen: Thanks for suggesting this!

I just experimented with a few different approaches for fitting this in, but none of them felt right.

I hope we communicate the "we care about performance" part well during the rest of the blog post. (As in "actions speak louder than words.")

Adding a paragraph before the CTA dilutes the CTA. It doesn't feel nearly as strong as before.

I think it's something we *should* communicate as a project, but I don't want to do it at this position in this blog post. (As you said, brevity can sometimes drive a message home better.)

We could make a social media post though with basically the content you just posted. That would be a great idea.

blog/tech-talk-window-resize-behavior.md
nikwen: Thanks for the suggestion! I thought about this.

The challenge is that resizing isn't smooth yet. It's just without those old pixels now. It's slow, much slower than in Chromium. (See https://github.com/electron/electron/issues/36280#issuecomment-3608288415.)

Therefore, a GIF wouldn't actually look that great.

blog/tech-talk-window-resize-behavior.md
nikwen: Thanks! I thought about this while writing the post. I decided not to make these headlines too descriptive because then they'd be spoilers for what was to come. That didn't feel right. (I do see the advantage in terms of the ability to skim.)

blog/tech-talk-window-resize-behavior.md
nikwen: Same as [for the other headline](https://github.com/electron/website/pull/958#discussion_r2636576460).

blog/tech-talk-window-resize-behavior.md
nikwen: Thanks for the suggestion! I put it in because I thought it was actually interesting to see that these kinds of issues are solved in collaboration with Chromium maintainers. I'd prefer to leave it in.

blog/tech-talk-window-resize-behavior.md
nikwen: Thanks for the suggestion!

Tbh I'd keep this as is.

* The v40 PR hasn't been merged yet. That's why I explicitly said the v39 PR was merged.
* I first didn't have the PR link here because of the repetition. But while re-reading it, I was like: "Where is the PR link?" It's not super obvious that the "Electron 39" link above goes to the PR. So I think it might be better to duplicate the link here.

Yes, we might be able to remove a sentence here, but tbh I don't have the motivation right now after editing the post based on review comments for 2.5 hours already.

PR #978, opened 2026-01-08
feat: transform non-docs relative links to GitHub content links

### LINKED ISSUE #977: Build relative links to `electron/electron` repo
In the website build process, it would be nice to be able to add relative links to the GitHub repo since the source of truth for the docs lives in https://github.com/electron/electron/tree/main/docs.

This could belong in an AST transformer or the `pre-build` script.

_Originally posted by @dsanders11 in https://github.com/electron/electron/pull/49294#pullrequestreview-3637234675_

dsanders11:
#### Description of Change

Closes #977.

It seems like by the time the transformer runs, all relative links that resolve to a file within `docs/latest` have already been turned into absolute paths starting with `docs/latest`, so really we could just transform any link that starts with `../`. To be safe though (in case that implicit behavior changes in the future), this PR does an additional check to confirm the relative path resolves to being outside of the `docs` directory before it transforms it into a GitHub link.

#### Checklist

### DIFF
--- a/src/transformers/github-content-links.ts
+++ b/src/transformers/github-content-links.ts
@@ -1,0 +1,65 @@
+import path from 'node:path';
+
+import { visitParents } from 'unist-util-visit-parents';
+import { Node, Parent } from 'unist';
+import type { Definition, Link } from 'mdast';
+import type { VFile } from 'vfile';
+import { latestElectronVersion } from '../util/latest-electron-version';
+import { isDefinition, isLink } from '../util/mdx-utils';
+
+const DOCS_FOLDER = path.join(__dirname, '..', '..', 'docs', 'latest');
+const RELATIVE_LINK_REGEX = /^(?:\.\.?\/)+(\S+)$/;
+
+let _version = '';
+async function getVersion() {
+  if (_version === '') {
+    _version = await latestElectronVersion();
+  }
+
+  return _version;
+}
+
+/**
+ * `attacher` runs once for the entire plugin's lifetime.
+ */
+export default function attacher() {
+  return transformer;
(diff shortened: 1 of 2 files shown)


src/transformers/github-content-links.ts
```diff
@@ -0,0 +1,65 @@
...
+ */
+async function transformer(tree: Parent, vfile: VFile) {
+  const version = await getVersion();
+
+  const findRelativeLinksOutsideDocs = (node: Node) => {
+    if (
+      (isLink(node) || isDefinition(node)) &&
+      (node.url.startsWith('./') || node.url.startsWith('../'))
+    ) {
+      // Check if the path resolves outside to be outside of the doc folder
```
erickzhao, line 41: Something I noticed is that relative links that try to go too far up the tree from the GitHub repo will resolve to a valid link.

```markdown
<!-- Will resolve to https://github.com/electron/electron/blob/v39.2.7/DEPS -->
[`DEPS`](../../../../../../../../../DEPS)
```

After double-checking, this seems to work on github.com's markdown rendering too so great!

erickzhao (approved):

src/transformers/github-content-links.ts
dsanders11: Yea, I took the approach that since `docs` is a top-level directory on `e/e`, any relative link going above it will immediately reach the top-level and can't go any further up, so it seemed safe to just strip all leading `../` segments without needing to do any extra logic.

The relative link linting on the `e/e` side should ensure we don't have any links which go too far up, but good to know that it will work even if one slips through. 👍

PR #981, opened 2026-01-16
fix(api-structure-previews): add cycle detection

erickzhao:
Due to changes in Electron 40.

### DIFF
--- a/src/transformers/api-structure-previews.ts
+++ b/src/transformers/api-structure-previews.ts
@@ -24,6 +24,37 @@ const fileContent = new Map<
   { promise: Promise<Parent>; resolve?: (value: Parent) => void }
 >();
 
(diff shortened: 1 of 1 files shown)


src/transformers/api-structure-previews.ts
```diff
@@ -24,6 +24,37 @@ const fileContent = new Map<
...
 >();
 
+// Tracks which files are waiting on which other files for cycle detection
+const waitingOn = new Map<string, Set<string>>();
+
+/**
+ * Detects if adding a dependency from `from` to `to` would create a cycle.
+ * Uses DFS to check if `to` transitively depends on `from`.
+ */
+function wouldCreateCycle(from: string, to: string): boolean {
```
erickzhao, line 34: Had to add cycle detection to this graph traversal 💀

PR #993, opened 2026-01-29
ci: gracefully handle no docs content to update

dsanders11:
#### Description of Change

As part of switching the website docs update to trigger off of new releases, there may be a release that doesn't include docs changes, so gracefully handle that case.

#### Checklist

### DIFF
--- a/.github/workflows/update-docs.yml
+++ b/.github/workflows/update-docs.yml
@@ -49,14 +49,14 @@ jobs:
             echo "changes=true" >> "$GITHUB_OUTPUT"
           fi
       - name: Push changes to branch
-        if: ${{ steps.prebuild.outputs.changes }}
+        if: ${{ steps.prebuild.outputs.changes == 'true' }}
         uses: dsanders11/github-app-commit-action@43de6da2f4d927e997c0784c7a0b61bd19ad6aac # v1.5.0
         with:
           message: 'chore: update ref to docs (🤖)'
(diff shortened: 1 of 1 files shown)


.github/workflows/update-docs.yml
```diff
@@ -38,20 +38,25 @@ jobs:
...
           if [ "$(git status --porcelain | grep -v 'docs/latest/.sha')" = "" ]; then
-            echo "Unexpectedly found no new content for docs - this is probably a bug"
-            exit 1
+            echo "::warning:: Unexpectedly found no new content for docs - this might be a bug"
+            echo "changes=false" >> "$GITHUB_OUTPUT"
+          else
+            echo "changes=true" >> "$GITHUB_OUTPUT"
           fi
       - name: Push changes to branch
+        if: ${{ steps.prebuild.outputs.changes }}
```
erikian: note: [this](https://github.com/orgs/community/discussions/163013#discussioncomment-13487333) seems to suggest that `false` will be treated as a string and not a boolean, so this condition is always fulfilled because `steps.prebuild.outputs.changes` is always truthy. Not sure if this changed recently, but we might need to explicitly compare this variable to `"true"` / `"false"`.

dsanders11: Yep, good catch! Updated it.

erikian (approved):

PR #1002, opened 2026-02-01
fix: fix outlined button styles and improve their accessibility

### LINKED ISSUE #1001: "See More" / "See Less" button on Governance page is buggy and potentially not accessible
https://github.com/user-attachments/assets/9ffbfdd6-05f4-4fa4-b211-0fb67ee275ec

Here I see two main issues, both sort of at odds:
1. Upon click (really focus) of the button the outline of the button is removed. I have tracked this down to the below CSS snippet that appears to have its origins upstream (EDIT: [the snippet appears to be from Docusaurus as of at least v3.9.2 (the current version we use)](https://github.com/facebook/docusaurus/blob/v3.9.2/packages/docusaurus-theme-common/src/hooks/styles.css)):
```css
body:not(.navigation-with-keyboard) :not(input):focus {
    outline: none
}
```
2. When using Tab-key navigation, when the button is selected and clickable, there is no visual indication (like there is for other elements) that it is selected. I believe because the outline it has by default is conflicting with the Tab-key-focus outline. This potentially makes the button less accessible to users who use Tab-key navigation.

EDIT:

Quick clarification: I just now realized that I had [Dark Reader](https://chromewebstore.google.com/detail/dark-reader/eimadpbcbfnmbkopoojfekhnkhdbieeh) enabled in the above screen recording. However, the behavior still persists with it turned off.

nmggithub:
#### Description of Change

Fixes #1001.

#### Checklist

- [X] This PR was not created with AI. (PRs created mainly with AI will be closed. They waste our team's time. We ban repeat offenders.)

### DIFF
--- a/src/css/custom.scss
+++ b/src/css/custom.scss
@@ -210,7 +210,8 @@ html {
   body:not(.navigation-with-keyboard) &--outline:focus {
     outline: 1px solid;
     color: var(--ifm-color-primary);
-    &:hover, &:focus-visible {
+    &:hover,
+    &:focus-visible {
       color: var(--ifm-color-primary-dark);
     }
   }


src/css/custom.scss
```diff
@@ -201,10 +201,14 @@ html {
 }
 
 .button {
-  &--outline {
+  &--outline,
+  // Docusaurus adds its own focus styles which we need to
+  // override here to avoid losing the outline.
+  // ref: https://github.com/electron/website/issues/1001
+  body:not(.navigation-with-keyboard) &--outline:focus {
```
nmggithub, line 210: This isn't ideal, but Docusaurus is too board in their approach to hiding focus outlines when the user is using a mouse (I added a comment here in this issue to get the attention of their team: https://github.com/facebook/docusaurus/issues/11314#issuecomment-3832188824)

nmggithub: The hook that requires this funky CSS by us has been removed upstream in Docusaurus (https://github.com/facebook/docusaurus/pull/11713), but no new version has been released yet.

PR #1125, opened 2026-05-18
build: drop tsx in favor of Node.js type stripping

dsanders11:
#### Description of Change

Main motivation here is the i18n deploy job has been failing since #1106, with some sort of JS import error.

To make this work we need to enforce file extensions (via `import/extensions` in oxlint) and enforce consistent usage of `type` imports (via `typescript/consistent-type-imports` in oxlint).

The helper scripts needed to be made ESM as part of this, so `import.meta.dirname` was swapped in for `__dirname` and that seems to be the only real change needed. There was also one minor tweak for importing from `@docusaurus/logger` to keep things happy with ESM, and I made it throughout the codebase to keep things consistent. The overall code base can't make the ESM leap until Docusaurus better supports it upstream, I think their goal is v4 will be the one.

There is one somewhat nasty hack in `scripts/prepare-i18n-content.ts` to make this CJS vs ESM dance work, unfortunately.

#### Checklist


ckerr (approved):

src/transformers/api-history.ts
```diff
@@ -1,13 +1,13 @@
...
+import { logger } from '@docusaurus/logger';
+import { ElectronVersions, type SemVer } from '@electron/fiddle-core';
 import AdmZip from 'adm-zip';
-import { Code, Node } from 'mdast';
+import type { Code, Node } from 'mdast';
 import semver from 'semver';
-import { Parent } from 'unist';
-import { ActionTuple, SKIP, visitParents } from 'unist-util-visit-parents';
+import type { Parent } from 'unist';
+import { type ActionTuple, SKIP, visitParents } from 'unist-util-visit-parents';
```
erikian, line 7: nit: since we're touching type-only imports, it might be a good opportunity to enforce one style consistently so we don't have both `import type { SomeType } from 'lib'` and `import { type SomeType } from 'lib'` scattered around the codebase.

I usually go with the latter (i.e. inline `type` imports / `'typescript/consistent-type-imports': ['error', { prefer: 'type-imports', fixStyle: 'inline-type-imports' }]`) as it allows mixing both type-only and regular imports in a single `import` statement and I find it easier to tell the type-only imports apart as they always have the `type` keyword right next to them (as opposed to the `type` keyword being present only once next to the `import` keyword in a type-only import statement), but either option would be fine as long as it's consistent.

dsanders11: Good point! When moving these over I was actually preferring inline if only a single import in a group was type only, and `import type` style if the full group is only type imports, to avoid redundancy. I can see the argument for consistency, though. Would be curious what the larger WG prefers, so we can enforce something consistent across our repos.

PR #1226, opened 2026-09-20
build: update @electron/lint-roller to 4.0.0

MarshallOfSound:
#### Description of Change

Bumps `@electron/lint-roller` to 4.0.0, which drops the transitive eslint/standard dependency tree; the shared markdownlint config and custom rules this repo uses are unchanged.

`lint-roller-markdown-links` no longer has a `--fetch-external-links` mode (electron/lint-roller#203, too many sites block automated fetches), so `check-blog-links.yml` drops that flag and keeps validating relative links and `static/` resources in new blog posts. While there, the step now uses `xargs -r` so a PR that only removes posts doesn't run the linter with no files and fail on its usage message.

oxlint stays at ^1.57.0: lint-roller 4 lists `oxlint >= 1.65` as a peer for its new `lint-roller-markdown-oxlint` command, which this repo doesn't use, and a newer oxlint flags a handful of React findings in `src/` that are better handled separately. Expect a peer warning on install until then.

#### Checklist

- [x] [I have reviewed and verified the changes](https://github.com/electron/governance/blob/main/policy/ai.md)

---
_Generated by [Claude Code](https://claude.ai/code/session_01GaskPz9P37WZvpQ5zKit7V)_


MarshallOfSound:
@SocketSecurity ignore npm/@electron/lint-roller@4.0.0

